Issue #19, opened 2017-09-22
Flat mode should be default in content selectors


alansemenov:
Flat mode should be default, unless overridden by ``config.treeMode = true`` (``config.flat`` should no longer be supported).

PR #26, opened 2017-09-22
Flat mode should be default in content selectors #19

vbradnitski:


alansemenov:
Pushed into branch "issue-24".

Issue #18, opened 2017-09-22
Translate Admin UI to Spanish


alansemenov:
@jsi commented on [Thu Sep 21 2017](https://github.com/enonic/xp-apps/issues/39)

Remember to include the common.properties file in lib-admin-ui also.

PR #31, opened 2017-09-25
Translate Admin UI to Spanish (#18)

aro:


aro:
@alansemenov There is a minor UI issue with the sign out text "Cerrar sesión", too long. I think it could be adjusted with CSS. Otherwise I could change it with "Cerrar" or try to find something else, but "Cerrar sesión" is what is used everywhere. 
Or maybe we should wait for the other language before doing size adjustments. For the rest I manage to make the text fit in the space for the XP tour, not so easy with Spanish.

sigdestad:
We should probably just fix css..

Issue #44, opened 2017-09-25
Missing images after cleanup


edloidas:
@sgauruseu commented on [Fri Sep 22 2017](https://github.com/enonic/xp-apps/issues/46)

1. Click on app in the Application grid
![apps](https://user-images.githubusercontent.com/3728712/30736615-21f9f458-9f8c-11e7-9174-dc8eccb2586a.png)

2. XP Tour
![image](https://user-images.githubusercontent.com/11612490/30743236-0e4d4f52-9f9d-11e7-8005-6a5f7a2d5161.png)

3. (Copied from https://github.com/enonic/xp-apps/issues/56) When "Insert Image" dialog opens from the HTML Area, there are errors in browser console. They were there even before the cleanup of images, but I don't see how these images are used in the dialog at all. Find out what's causing the errors.
![image](https://user-images.githubusercontent.com/11612490/30806066-8b12abbe-a1f4-11e7-890e-4979a265fa1d.png)

PR #47, opened 2017-09-25
Missing images after cleanup #44

edloidas:
Fixed the missing arrows in image modal dialog.


src/main/resources/assets/admin/common/styles/api/util/htmlarea/modal-dialog.less
```diff
@@ -310,10 +312,16 @@
...
+        background-image: url('../../../../images/arrow_up.png');
+      }
     }
 
     .scroll-down-div {
       bottom: 0;
+      .arrow {
+        background-image: url('../../../../images/arrow_down.png');
+      }
     }
```
alansemenov: We shouldn't need ``.arrow``, ``.scroll-up-div`` and ``.scroll-down-div`` in ``modal-dialog.less``. The arrows are specific for the Insert Image dialog, so these classes should only be in ``image-dialog.less``

Issue #20, opened 2017-09-22
Add toggle tree icon to content selector


alansemenov:
Replace the``input-icon`` on the left side of the selector which is not doing anything with the ``tree`` icon that will toggle tree mode on and off.

Replace
![image](https://user-images.githubusercontent.com/11612490/30739724-ae8c6318-9f8e-11e7-99ae-25dc76cb3e23.png)
with
![image](https://user-images.githubusercontent.com/11612490/30739805-f3aa2264-9f8e-11e7-93aa-98094bf43033.png)

The icon's toggle state must be based on whether the selector is currently in tree mode or not.
If selector's config has ``treeMode=true``, then the icon should be by default toggled ON.

Tooltips (must be localized):
When the icon is "ON": _Switch to flat mode_
When "OFF": _Switch to tree mode_

PR #50, opened 2017-09-25
Add toggle tree icon to content selector #20

vbradnitski:


alansemenov:
@vbradnitski it's looking really good, but I assume it's still work in progress because the flat mode still has margin on the left side and expand icon for nodes (which gives an error when clicked). But this task was only for adding the icon, so I merged it into branch "_issue-24_" (this is the branch for the epic which all pull requests should target, we'll merge when everything is ready).

image

alansemenov:
Also LoaderCombobox (Applications, Language and Owner selectors in the Site Wizard) doesn't seem to be working at all now in the branch _issue-24_ - the dropdown is empty, no errors in the console.

Issue #74, opened 2017-10-04
IE11: Content Studio is blank


ashklianko:
Error in IE11 occurs, CS remains blank when loading.

PR #75, opened 2017-10-04
IE11: Content Studio is blank #74

ashklianko:
-IE11 does not support iteration features of Map, in particular values() and keys() trigger errors in IE11; replaced with valid alternatives

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/ValidationRecording.ts
+++ b/src/main/resources/assets/admin/common/js/form/ValidationRecording.ts
@@ -1,6 +1,6 @@
 module api.form {
 
-    import ArrayHelper = api.util.ArrayHelper;
+
 
     export class ValidationRecording {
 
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/form/ValidationRecording.ts
```diff
@@ -168,23 +168,14 @@ module api.form {
...
-                    const key: string = result.value;
-
-                    if (!mapA.get(key).equals(mapB.get(key))) {
-                        return false;
-                    }
-                }
-
-            } while (!result.done);
+            const keys: string[] = [];
+            mapA.forEach((value: KeyBinding, key: string) => {
```
aro: Why is the type of value here KeyBinding? 
According to function signature it should be Equitable. I'm surprised TypeScript compiler does not complain.

Issue #111, opened 2017-10-18
Browse Panel - Remove support of tablet resolution mode


alansemenov:
We now support three layout modes in the browse panel of all the three main applications: desktop (preview/stats panel on the right), tablet (preview panel at the bottom) and mobile (slide-in preview panel). We will ditch the tablet mode and only support the other two. Preview will always be on the right except for mobile mode where it will be hidden by default and slide-in on content item click.

alansemenov:
Have to fix the detail panel behaviour (see comment to PR)

PR #125, opened 2017-10-20
Browse Panel - Ditch tablet mode #111

pmi:


alansemenov:
@pmi Works fine, but now when the page is in tablet width and detail panel is expanded, there's not much space left for the preview (see below). We should change it so that detail panel goes in floating mode on widths > mobile and < tablet (basically same as when we used to switch preview to be under the grid).

image

sigdestad:
Hmm. Looking at Alans screenshot. I think what is wrong is that there should be a minimum width for the "grid" as long as we are operating in a split screen mode. Can we try "freezing" at a minimum of 300px before we make changes to the preview panel. We carefully crafted how the detail panel should work in these cases, and I would like to move carefully when changing these things.

sigdestad:
So to followup - For the screen resolution above we should probably already use mobile mode?

alansemenov:
> For the screen resolution above we should probably already use mobile mode?

I'm not so sure. 768px width is enough for both the grid and the preview side by side, it's only when the details panel expands when the preview panel gets squeezed between that and the grid.

Grid width is already set to 300px in tablet mode (with remaining 468 for the preview panel).

sigdestad:
Hmm.. I thought we said to use mobile mode for <960px?

alansemenov:
This is what we have now:

Mobile (slide-in preview): width < 540px
Tablet (preview at the bottom): 540px < width < 720px
Desktop (preview to the right): width > 720px

After removal of the tablet mode it's mobile mode under 540 and desktop over 540.

A threshold of 960px is used only in the Wizard to determine whether the Page Editor should be visible by default or not.

Should we say that anything under 960px is "mobile" mode with the right panel hidden by default in both the Grid and the Wizard?

What I don't like in mobile preview is that you cannot see both the grid and the preview at the same time. While it's necessary on low resolutions (under 540) it doesn't make much sense on, say, 959px - there's plenty of room to show both the grid and the preview side by side. Maybe we should use 720px as mobile/desktop threshold?

sigdestad:
Lets try 720 then!

Issue #129, opened 2017-10-23
Implement image preview in Image Selector in desktop resolution


alansemenov:
When user hovers mouse cursor over a thumbnail we should expand a bigger version of image next to the cursor. The trick here is to find a position of the preview placeholder based on the cursor position, f.e. if thumbnail is closer to the left border then the preview should open to the right, and vice versa. The preview should automatically disappear on mouse out. Left click on the thumbnail image should select the option, as before.

It should work almost like here: https://www.magictoolbox.com/magiczoomplus/ (hover over the bag image) except that we don't need to zoom into parts of the original image - we will always show the entire (but bigger) image in the preview. Note the transition effect when the preview kind of slides to the right from the original image - we should try to implement something like that too. 

Also note that when there's not enough space to show the original image and the preview side by side then the preview will be shown over the original image.

I suggest dynamically creating the image preview div with absolute positioning and fixed width (height should be calculated based on width) on thumbnail hover, and dynamically destroying it on mouse out.

alansemenov:
Some issues described in the PR

PR #135, opened 2017-10-26
Implement image preview in Image Selector in desktop resolution #129

pmi:
- automatically choses side of the picture with which has more space
- preview scales to occupy all available size as long as there are more than 500px available
- preview doesn't shrink less than 500px and overlays the thumb in that case
- it also doesn't go beyond page boundaries and shows scrollbars in that case

### DIFF
--- a/src/main/resources/assets/admin/common/js/content/image/ImageContentComboBox.ts
+++ b/src/main/resources/assets/admin/common/js/content/image/ImageContentComboBox.ts
@@ -5,6 +5,7 @@ module api.content.image {
     import ContentTypeName = api.schema.content.ContentTypeName;
     import OptionDataHelper = api.ui.selector.OptionDataHelper;
     import ComboBox = api.ui.selector.combobox.ComboBox;
+    import ElementHelper = api.dom.ElementHelper;
 
     export class ImageContentComboBox
         extends ContentComboBox<ImageTreeSelectorItem> {
@@ -18,7 +19,28 @@ module api.content.image {
 
             super(builder);
 
-            this.addClass('image-combo-box');
+            this.addClass('image-combobox');
+
+            this.initImagePreview();
+        }
+
+        private initImagePreview() {
+            const preview = new ImageContentPreview();
+
+            this.onMouseOver((e: MouseEvent) => {
+                const helper = new ElementHelper((<HTMLElement>e.target));
+                if (this.isThumbnailImage(helper)) {
+                    preview.showFor(helper);
+                }
+            });
+
+            preview.onSelected((id: string) => {
+                this.selectOptionByValue(id);
+            });
+        }
+
+        private isThumbnailImage(helper: ElementHelper): boolean {
+            return helper.getTagName().toLowerCase() == 'img' && helper.hasAttribute('data-contentid');
         }
 
         getContent(contentId: ContentId): ContentSummary {
--- a/src/main/resources/assets/admin/common/js/content/image/ImageContentPreview.ts
+++ b/src/main/resources/assets/admin/common/js/content/image/ImageContentPreview.ts
@@ -12,7 +12,7 @@ module api.content.image {
         private static PADDING: number = 20;
         private static TIMEOUT: number = 300;
 
-        public static debug: boolean = true;
+        public static debug: boolean = false;
 
         private image: ImgEl;
 
@@ -31,8 +31,10 @@ module api.content.image {
             this.resolver = new ContentImageUrlResolver();
             this.image = new ImgEl();
             this.image.onLoaded(() => {
-                this.mask.hide();
                 if (!this.image.isPlaceholder()) {
+                    this.mask.hide();
+                }
+                if (this.isVisible()) {
                     this.positionFor(this.thumb);
                 }
             });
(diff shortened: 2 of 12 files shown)


sigdestad:
Do we have a screenshot of this in action?

alansemenov:
Some issues (feel free to create sub-issues under the epic since there are quite a few):
* The preview is triggered even on content items that are not images (try the tree-structured image selector where we show parent nodes of images). That was the reason for my initial commit into the "issue-128" branch where only image content items get the "image" CSS class applied. The idea with contentId inside a data- attribute is good, but then it must be set only on images.
* Both scrollbars are always visible on the image preview - there should never be any scrollbars in the preview
* We need an unobtrusive border around the preview - images with white background are not looking very good on the wizard's white background
* When the preview is on and the thumbnail gets clicked it takes a second or two before the preview fades out. In this case we don't need this effect, the preview should disappear right away.  But we should still have it when cursor hovers from one thumbnail to another. 
* Image preview of an Image component inside Inspect panel of the Page Editor is a bit buggy - preview of some images inside the Features site opens over the selector even though there's plenty of space to the left of it:

image

* We should constrain the image preview to be inside the available area. Some of the (bigger) images are going way out of the available viewport. So it's not only width that should be considered, but also height.

image

* Sometimes ``size`` parameter of the image url is calculated with decimal part - in this case preview will fail

image

* When image selector is inside a modal dialog (site config, Insert Image dialog of HTML Area) then preview will be covered by the body shader. It should have a higher z-index to open over it. 

image

sigdestad:
Looks like you were missing a design for this task Pavel?

We need this to work more naturally in our current UI - also I guess the mobile and "desktop" versions could work pretty much the same. Attached is my suggestion to solve this:

* If preview area is available, simply use this to show the preview - even without the user having to click anyhere. It is enough to use keyboard or mouse-over like in the suggestion below
* If no preview is available (hidden or we are in mobile mode), the user must tap/click the image thumbnail, and the same preview should appear, but this time fill the screen. 

There needs to be a "max" size for the image based on the smallest part of height/width and also reserve some space for select button etc.

Notice: When preview is active, the up/down buttons should still work like today moving up/down is still working inside the content/image selector and now also refreshing the preview. Pressing space (in bulk) should still select content, and pressing enter "select + apply" - so no changes except we get a big preview. Pressing esc button should exit naturally in both cases I guess.

image-zoom 003
image-zoom 002
image-zoom 001

NB! I did not look at Alans mobile suggestion, but I think this is similar to what he designed there?

sigdestad:
NB! I suggest using a grey 10% transparent background for this design - not the standard modal black...

sigdestad:
Lets conclude the design before any more coding is done please :-)

sigdestad:
Also, for modal dialogues, we should always use the "no preview available" mode I guess...

pmi:
I actually like the idea of reusing the preview area for showing an image! It fits the general application appearance better.
But clicking an image in a desktop mode is not a good idea because clicking a list item naturally selects it.
What we could do instead is showing a zoom icon above thumb on hover in all modes, and then showing a preview popup (same as the mobile one) upon clicking it in case there is no preview area available or showing an image in preview area on hover otherwise.

sigdestad:
Sound interesting Pavel. Could you create a sample screenshot of how this would look? @ase do you have any ideas around this? Anything that would indicate clicking the image triggers a special action would be nice I guess - even without the hover maybe?

alansemenov:
> @ase do you have any ideas around this? Anything that would indicate clicking the image triggers a special action would be nice I guess - even without the hover maybe?

Well, I definitely don't like the idea of having to click the thumbnail to get the preview and then click again to close it. User may have to preview tens of images to find the one to select and he will have to click twice on each? The idea was to quickly get the preview up by hovering cursor over a thumbnail and then quickly close it by hovering out or over another image.

Using Page Editor panel for an image preview is ok I guess, but that's just one case out of many where we need to show the preview - wizard without the preview, wizard with collapsed editor panel, image selector inside the Inspect panel, image selector inside the Page Editor, image selector inside the modal dialog - so we need one consistent solution that will cover all cases, and I don't really see how the new suggestion helps here?

My idea was to have something simple like here: http://cssglobe.com/lab/tooltip/02/ and I'm not sure why it has to be more complicated than that except on mobile where we can't use hover.

sigdestad:
Good point indeed Alan, the preview panel idea sounds a bit like a hack. I had actually forgotten about context panel for a second. I really don't like the hover solution - as this then gets in the way of other info. Maybe the best solution would simply be applying a grid mode for imageSelector instead? We still get issues with tree-navigation though? or do we?

sigdestad:
Here is a different approach to zooming. So on-mouse-over would magnify the entire row and show a square cropped version of the image? BTW: For desktop mode I definetly think the Apples "click space for preview" approach is really nice, but then we would have to do something different for bulk selection button?
screen shot 2017-10-31 at 07 47 21

sigdestad:
Also, another solution would be to maximize use of the current available space, and rather pan the image if it is not square?

sigdestad:
Maybe a combination of these things would be nice:
* Mouse-over/focus on row maximizes the icon within the current borders (the square trick)
* Click on icon reveals the "fullscreen" preview. We also need a keyboard shortcut for this.. Space?

sigdestad:
Also.. For the image preview feature I would like to directly re-use the preview functionality we have in the grid view. This works very nicely in terms of image size etc.

Issue #85, opened 2017-10-09
Fix build problems on Concourse for lib-admin-ui.


jsi:
There is something that is cached incorrectly, that causes some dependencies to be left out when Concourse builds the library.

jsi:
The main problems turns out to be that _all.js is not built and included.  An attempt to do add "install" as a task for the gradle build, did not help any.

jsi:
The file "_all.js.map" is copied as before, with one build delay, so local builds include the previous version, while the build server does not have it.  This file is not needed, and using a version of lib-admin-ui without it, produces a distro that is about 3Mb smaller, so that should be the preferred build.

PR #136, opened 2017-10-26
Fix build problems on Concourse for lib-admin-ui. #85

edloidas:
Fixed the problem of the compiled JS not present under the build folder, since `:processResources` task executes before the `gulpAll`.

**Comments on the problem:**
_When you run `gradle clean` locally it will only remove the `build/` folder. Our compiled files are placed under the `src/main/resources/**/common/js/` folder. They won't be removed with clean build. So the problem will occur even on local machine, if you have run the `gradle build` for the first time on newly cloned repo.
And the problem is that the `:processResources` task that copies all the source resources to `build/resources` runs before the `gulpAll` task, that compiles `.ts` and libs to `_all.js` files._

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,16 @@ task gulpAll( type: GulpTask, dependsOn: npmInstall ) {
     args = ['all']
 }
 
-task copyTypeDefs( type: Copy, dependsOn: gulpAll ) {
+task copyCompiled( type: Copy, dependsOn: gulpAll ) {
+    description = 'Copy compiled js files.'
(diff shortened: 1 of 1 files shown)


alansemenov:
Nice! @jsi please verify that this works.

sigdestad:
Nice job @edloidas

jsi:
These three files are still not copied over:
```
./build/resources/main/assets/admin/common/styles/_all.css
./build/resources/main/assets/spec/_all.js
./build/resources/main/assets/spec/_all.js.map
```

PR #137, opened 2017-10-28
Add update of the `_all.css` file in the build folder

edloidas:


alansemenov:
Hmm, so this has never worked? How did _all.css get to the xp-apps before?

edloidas:
@alansemenov Same way the `_all.js` did. I've described it in the previous [PR](https://github.com/enonic/lib-admin-ui/pull/136). 
__TLDR__: `:processResources` Gradle task runs before all other tasks. It was the main task, that copies resources. Last time I've added `copyCompiled` task, that copies only compiled files to `build/` folder. I thought CSS was copied but was wrong. As @jsi [mentioned](https://github.com/enonic/lib-admin-ui/pull/136#issuecomment-339901822) CSS was missing on the first build because it wasn't yet created, when `:processResources` started.
As for the files in `spec/` folder, we exclude them from `.jar` on purpose, since there are front-end tests only.

edloidas:
But it only concerns the first-time build. Also, don't forget, that when the server is running in dev mode, it will load the data from `src/` folder, not from `build/` or from `.jar` itself.

alansemenov:
Merged. @jsi please verify

Issue #188, opened 2017-11-17
Row height in Image Selector's gallery mode must be responsive


alansemenov:
Gallery mode looks fine in Image Selector when it has default width of over 500px but when it's smaller, the images start looking weird (see below).

240px (Inspect panel)
![image](https://user-images.githubusercontent.com/11612490/32948923-3ae3bcd8-cba1-11e7-8227-54ea50bc9967.png)

320px (Content Wizard in mobile mode)
![image](https://user-images.githubusercontent.com/11612490/32948769-b419a4b0-cba0-11e7-9c55-d3811f0e5276.png)

Size of the image themselves is fine - they are correctly resized according to the selector width - but we need to correctly calculate *height* of the row element based on the grid width.

For example, the selector in Inspect panel should look like this:
![image](https://user-images.githubusercontent.com/11612490/32949201-42e5987e-cba2-11e7-9685-98eb93b76108.png)

Row height must be responsive and recalculate on every change of the grid's width (but only in gallery mode).

PR #204, opened 2017-11-21
Row height in Image Selector's gallery mode must be responsive #188

pmi:


alansemenov:
It works fine but there's a new bug introduced after the changes. When you add a new Image component to the page, it gets expanded right away and - because load is triggered by the click on the toggle button - the dropdown is empty. It shouldn't expand.

image

alansemenov:
We should also set font-size of image's display name relative to the available width - I tried with ``font-size: 1.5vw`` and it seems to work nicely.

pmi:
Keep forgetting page editor has it's own styles -- fixed that now.

Regarding *vw*, it's relative to viewport width, so it will grow constantly with the viewport.
But images size will only grow along with the combobox width, that is limited to ~600px.

Issue #205, opened 2017-11-21
Give warning when trying to delete content with inbound dependencies


vbradnitski:
@alansemenov commented on [Tue Nov 14 2017](https://github.com/enonic/xp-apps/issues/258)

Deleting a content that other content items have references to can make them unrenderable, so we should give a warning to a user  trying to delete such items. We already have a Dependencies widget where we show Incoming and Outgoing dependencies, and it's incoming dependencies which we are interested in.

![image](https://user-images.githubusercontent.com/11612490/32776047-97d56094-c931-11e7-95fc-1b296c57fa25.png)

In the first revision we will not check inbound dependencies of **dependant** items, but **only the main (selected) ones**.

1. When we show DeleteContent dialog, make a request to fetch inbound dependencies of **all** of the selected items **at once**. If we have paging in the list of main items it should not affect the request - we should always fetch dependencies for the entire list.

**Note**: we need to change getDependencies REST API from GET to POST, pass multiple IDs and receive an object with inbound and outbound dependencies for each passed Id: https://github.com/enonic/xp/issues/5784

2. If any of the selected items has at least one inbound dependency, show a warning (docked, not auto-hide):

_The content you are about to delete has inbound references. Please verify them before deletion._

![image](https://user-images.githubusercontent.com/11612490/32777771-bf10db9c-c937-11e7-99e8-210f62ae9fc6.png)

3. Automatically close the warning when DeleteContent dialog is closed.

4. For each item that has inbound dependencies, show their number under the item. Use singular resource for 1 item and plural for >1.

![image](https://user-images.githubusercontent.com/11612490/32776026-853bde22-c931-11e7-9a20-6068574f8d97.png)

5. The link should be clickable and open the list of inbound dependencies - same as what happens when clicking "Show inbound" button in the Dependencies widget - but **in a new tab**. Every link should open in a new tab, not all of them in the same, so use content Id as window Id.

![image](https://user-images.githubusercontent.com/11612490/32777832-0b73c04e-c938-11e7-9e0a-4c204d7152a3.png)

6. Localise text of the warning and "_Inbound dependency_" / "_Inbound dependencies_".

7. Make sure that we fetch current list of dependencies each time the DeleteContent dialog opens, so that if I removed inbound dependencies in other browser tabs and then reopened the Delete dialog the deleted dependencies won't be there.

PR #206, opened 2017-11-21
Give warning when trying to delete content with inbound dependencies …

vbradnitski:
…#205

### DIFF
--- a/src/main/resources/assets/admin/common/js/app/AppPanel.ts
+++ b/src/main/resources/assets/admin/common/js/app/AppPanel.ts
@@ -1,6 +1,7 @@
 module api.app {
 
-    export class AppPanel<M extends api.Equitable> extends api.ui.panel.DeckPanel {
+    export class AppPanel<M extends api.Equitable>
+        extends api.ui.panel.DeckPanel {
 
         protected browsePanel: api.app.browse.BrowsePanel<M>;
 
--- a/src/main/resources/assets/admin/common/js/app/bar/AppBarTabId.ts
+++ b/src/main/resources/assets/admin/common/js/app/bar/AppBarTabId.ts
@@ -27,9 +27,8 @@ module api.app.bar {
             return this.id;
         }
 
-        changeToEditMode(id: string) {
-            this.mode = 'edit';
-            this.id = id;
+        getMode(): string {
+            return this.mode;
         }
 
         equals(other: AppBarTabId): boolean {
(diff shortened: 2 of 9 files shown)


alansemenov:
It works fine, but we need to optimize the code - there are too many unnecessary loops we should avoid.
First thing we need to do is build a "<contentId>: <count>" structure where we aggregate number of incoming dependencies for each source contentId. This should be done in ResolveDependenciesResult:
```Javascript

        public getIncomingDependenciesCount(): Object {
            let object = {};
            this.dependencies.forEach(dependencyResult => {
                const dependency = dependencyResult.getDependency();
                const contentId = dependencyResult.getContentId().toString();
                if (dependency.inbound && dependency.inbound[0]) {
                    object[contentId] = dependency.inbound.reduce((sum, dep) => sum + dep.count, 0);
                }
            });

            return object;
        }
```

This method will return the object looking like this:
```JSON
{
  "contentId_1": 5,
  "contentId_2": 1,
  "contentId_N": 2
}
```
and it will only calculate number of inbound dependencies, so we don't have to run any ``filter`` anywhere else.

Next, we need to use this object from inside the ContentDeleteDialog:

```Javascript
    protected manageInboundDependencies(contents: ContentSummaryAndCompareStatus[]) {
        new api.content.resource.ResolveDependenciesRequest(contents.map(content => content.getContentId())).sendAndParse().then(
            (result: api.content.resource.ResolveDependenciesResult) => {

                const dependencyCount = result.getIncomingDependenciesCount();

                if (!Object.keys(dependencyCount).length) {
                    return;
                }

                this.messageId = api.notify.showWarning(
                    i18n('dialog.delete.dependency.warning'), false);

                this.getItemList().getItemViews().forEach((itemView) => {
                    const contentId = itemView.getBrowseItem().getModel().getContentId().toString();

                    if (dependencyCount.hasOwnProperty(contentId)) {
                        (<DeleteItemViewer>itemView.getViewer()).setInboundDependencyCount(dependencyCount[contentId]);
                    }

                });
            });
    }
```

This is more readable and will execute much faster. 

Also, there are a couple of nasty issues when we loop through all item views even when there are no dependencies for any items, and when the AEl element was created in DOM even when it was empty. The last one I fixed directly in the PR.

alansemenov:
...and I guess we won't need ``ResolveDependenciesResult.getById`` after these changes.

Issue #257, opened 2017-12-15
A click on warning about incoming dependencies closes the Delete dialog


vbradnitski:
@alansemenov commented on [Fri Dec 15 2017](https://github.com/enonic/xp-apps/issues/375)

When you're about to delete content that has incoming dependencies, we show a warning about this when opening the dialog. This is fine, but if I want to close the warning and click on X then the dialog recognises this as a click outside of the dialog and closes itself.

We should allow a click on the notification and not close any modal dialog after that.

PR #258, opened 2017-12-15
A click on warning about incoming dependencies closes the Delete dial…

vbradnitski:
…og #257

### DIFF
--- a/src/main/resources/assets/admin/common/js/notify/NotifyManager.ts
+++ b/src/main/resources/assets/admin/common/js/notify/NotifyManager.ts
@@ -75,7 +75,7 @@ module api.notify {
             const registryArray = Object.keys(this.registry).map((key) => this.registry[key].opts);
 
             return registryArray.some((registryEntry) =>
-                                            registryEntry.message == opts.message && registryEntry.type == opts.type);
+                registryEntry.message == opts.message && registryEntry.type == opts.type);
         }
 
         private createNotification(opts: NotifyOpts): NotificationMessage {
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/notify/NotifyManager.ts
```diff
@@ -93,6 +93,12 @@ module api.notify {
             return notificationEl;
         }
 
+        getNotification(messageId: string): NotificationMessage {
+            if (messageId && this.registry[messageId]) {
+                return this.registry[messageId].el;
+            }
+        }
```
alansemenov: A non-void method should always return something. ``getNotification`` should return ``null`` if the notification doesn't exist.

src/main/resources/assets/admin/common/js/ui/dialog/ModalDialog.ts
```diff
@@ -208,6 +208,12 @@ module api.ui.dialog {
             this.listOfClickIgnoredElements.push(elem);
         }
 
+        removeClickIgnoredEl(elem: api.dom.Element) {
+            if(this.listOfClickIgnoredElements.indexOf(elem) >= 0) {
+                delete this.listOfClickIgnoredElements.splice(this.listOfClickIgnoredElements.indexOf(elem), 1);
+            }
+        }
```
alansemenov: Assign ``this.listOfClickIgnoredElements.indexOf(elem)`` to a ``const`` and use it the second time inside the `ìf``.

PR #260, opened 2017-12-18
UserWizard - trim spaces in a password field #259

edloidas:
Added validation of the password filed by symbols.


src/main/resources/assets/admin/common/js/ui/text/PasswordGenerator.ts
```diff
@@ -120,22 +120,31 @@ module api.ui.text {
...
-            } else if (this.isWeak(value)) {
-                this.complexity = 'weak';
+                this.complexity = null;
             }
-            if (this.complexity) {
-                this.addClass(this.complexity);
-                this.getEl().setAttribute('data-i18n', i18n(`field.pswGenerator.complexity.${this.complexity}`));
+
+            this.toggleClass('invalid', !isValid);
+
```
alansemenov: Shouldn't we use ``FormItem.INVALID_CLASS`` here as well?

edloidas: We have a plenty of places, where we used `'invalid'` as string literals in the descendants of the `FormItem`. We should do the refactoring there, but only after the release.

Issue #269, opened 2017-12-20
Page Inspection panel - Template name is missing for the Automatic template option


pmi:
@sgauruseu commented on [Wed Dec 13 2017](https://github.com/enonic/xp-apps/issues/359)

1. Open Features
2. Open the Page Component View and click on 'customize page'
3. Open the Inspection Panel and click on dropdown-handler  in PageTemplateSelector

BUG: Automatic option displayed with the message: 'no default template found', but actually , 'main' template should be present in options

![def-templ](https://user-images.githubusercontent.com/3728712/33930638-aa0ef7be-dffe-11e7-9b57-680797f29585.png)

PR #271, opened 2017-12-20
Page Inspection Panel - default template is not present in the option…

pmi:
…s of PageTemplateSelector #269

Page Inspection Panel - when a template has been removed it should not be displayed in the options of PageTemplateSelector xp-apps#360


pmi:
Handled all sane cases of template manipulation in live edit:

- templates sorted
- template updated
- template added/removed
- supports added/removed
- switch to automatic mode
- switch to no controller mode

alansemenov:
Great job, @pmi

pmi:
Muchas gracias, senor ;)

Issue #290, opened 2018-01-08
Instantly invalidate new content with invalid default value


vbradnitski:
@sgauruseu commented on [Thu Dec 28 2017](https://github.com/enonic/xp-apps/issues/425)

1) Add the next content type:
`<input name="socialSecurityNumber" type="TextLine">`
`<occurrences minimum="0" maximum="0"/>`
`<config> `
` <max-length>11</max-length>`
`</config> `
`default>000-00-000000</default>`
2. Open a wizard for the content and type a name and press the Save button

<img width="882" alt="textline2" src="https://user-images.githubusercontent.com/3728712/34406180-a1fd5220-ebc8-11e7-8846-0d1428508228.png">

---

@alansemenov commented on [Tue Jan 02 2018](https://github.com/enonic/xp-apps/issues/425#issuecomment-354762056)

We should validate default value as we would validate any value entered in the field. If the field with default value is not valid it should be instantly highlighted, just like it happens invalid value is entered manually. In this case invalid value will not be sent to server when saving content.

PR #291, opened 2018-01-09
Instantly invalidate new content with invalid default value #290

vbradnitski:


alansemenov:
Seems to be working fine

Issue #340, opened 2018-02-15
Misc improvements and bugs in Publishing issue dialog


alansemenov:
1. Selecting "Edit" on a comment should set focus inside the comment field, clear and disable the new comment field and disable the buttons
2. Pressing "Esc" inside the new comment should clear and blur the new comment field. Now it closes the dialog and takes the user back to the issue list.
3. Bug: the new comment field won't get cleared when switching between different issues
4. Bug: when I press a key inside the comment being edited, the Save button won't get enabled. Only on the second key press it becomes enabled.
5. It should be possible to delete the last item on the Items tab (now the "x" icon is disabled).

alansemenov:
@pmi See comments inside the PR

alansemenov:
One more thing. Please remove the description from the email - we show it as the first comment and now it's just duplicating info. After that make sure we have the same margin above and under the "Show details" button.

![image](https://user-images.githubusercontent.com/11612490/36535071-c00df87e-17c8-11e8-9b89-27b0b32d9953.png)

PR #346, opened 2018-02-20
Misc improvements and bugs in Publishing issue dialog #340

pmi:


alansemenov:
These two are not fixed:

> Selecting "Edit" on a comment should set focus inside the comment field, **clear and disable the new comment field** and disable the buttons

The buttons get disabled but the new comment field is not cleared and not disabled. We should either clear it or disable it.

> Bug: when I press a key inside the comment being edited, the Save button won't get enabled. Only on the second key press it becomes enabled.

Still have to press the key twice before the button gets enabled.

pmi:
@alansemenov Text area is actually disabled using readonly attribute. I do not advise to clear it because the user might have typed in something and will loose it otherwise. 
Double key press is also fixed, at least on windows ;)

alansemenov:
The problem with textarea is that it doesn't look disabled at all, and it's confusing to see two textareas full of text, both looking editable. We should either apply disabled attribute or tweak opacity in CSS to make the New comment textarea really look disabled.

Issue #374, opened 2018-02-28
Site Wizard - Invalid application is not highlighted after version rollback


edloidas:
@sgauruseu commented on [Tue Jan 30 2018](https://github.com/enonic/xp-apps/issues/520)

1. Open the site wizard and start to add a new site.
2. Type a name and select 'Google Analytics App'
3. The red icon appears in the wizard because the required field is not filled in the site-configurator. 
   And the selected option view is red(OK)
![step1](https://user-images.githubusercontent.com/3728712/35562947-61cac118-05b5-11e8-9afc-e4593ca45691.png)

4. Click on the `Edit ` icon and open the site-configurator, type a required data and press the `Apply`
 The site is getting valid and the selected option is not red now
5. Do not close the site-wizard and go to the grid, open the Version History panel and restore the previous version.
6. Go to the wizard again.

__Expected:__ selected  application should be red
__Actual:__ option is not red, but the required input is empty

PR #375, opened 2018-02-28
Site Wizard - the color of selected application-view is not updated, …

edloidas:
…when switch between versions and site configurator in an application contains required fields #374

* Fixed `SiteConfigurator` update method by adding the missing update with the new `PropertyArray` each of the selected properties.


edloidas:
It seems case 2 have appeared again, so this PR will fix only the first problem.

alansemenov:
@edloidas Are you going to fix the second case in the same pull request or should we register a separate bug for it?

edloidas:
@alansemenov I'll move it to the separate issue.

PR #389, opened 2018-03-06
Site Wizard - The list of selected applications is not refreshed after version rollback

alansemenov:
Added combobox selected options update from site configurator on `propertyArray` update.
Set `ignorePropertyChange` to `true` before updates to avoid cyclic `update` method calls, when multiple options are selected and updated.

### DIFF
--- a/src/main/resources/assets/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfigurator.ts
+++ b/src/main/resources/assets/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfigurator.ts
@@ -1,4 +1,5 @@
 module api.content.site.inputtype.siteconfigurator {
+    import Property = api.data.Property;
     import PropertyArray = api.data.PropertyArray;
     import FormView = api.form.FormView;
     import Value = api.data.Value;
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfigurator.ts
```diff
@@ -74,12 +74,38 @@ module api.content.site.inputtype.siteconfigurator {
...
+                    const view = <SiteConfiguratorSelectedOptionView>this.comboBox.getSelectedOptionByValue(key).getOptionView();
+                    const configSet = set.getProperty('config').getPropertySet();
+                    const update = view.getFormView().update(configSet, unchangedOnly);
+                    updatePromises.push(update);
+
+                    optionsToRemove = optionsToRemove.filter(option => optionToKey(option) !== key);
                 });
+
+                optionsToRemove.forEach(option => this.comboBox.deselect(option.getApplication(), true));
+
```
alansemenov: @edloidas We need to refactor this new code, as in extracting it to new method(s) with self-explanatory name(s). It's very unobvious now what the code does and why it's suddenly so many of it.

Issue #413, opened 2018-03-27
Gap between input and dropdown inside the descriptor selector


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/37965489-8e8a232e-31c5-11e8-9bd8-b99ed5064b27.png)

PR #421, opened 2018-04-02
Gap between input and dropdown inside the controller selector for a n…

vbradnitski:
…ew site #413


vbradnitski:
jQuery 3.2 version bug

PR #446, opened 2018-04-18
Implement integration with CKEditor

ashklianko:


alansemenov:
@ashklianko  Instead of having checks like ``if this.isCKEditor...`` all over the HtmlArea class, we need to implement an entirely new class for the CKEditor (call it for example, ``HtmlAreaCK``) and create an instance of *that* class if the config option is turned on. This way we won't have to go through the entire HtmlArea class and remove the checks when it's time to drop TinyMce, we'll just rename ``HtmlAreaCK`` into ``HtmlArea`` replacing the old class with TinyMce integration.

You can also work with this new class without having to worry about breaking current integration with TinyMCE.

ashklianko:
Yey! :)

Issue #461, opened 2018-04-26
HtmlArea content with CKE - incorrect behavior of validation, when required inputs in wizard 


sgauruseu:
1. open new wizard for htmlarea 2:4(two required)
2. Type a name and type text in both areas
3. Save the content(do not close)
4. Delete text in areas and save the content

Bug required area is empty, but red icon does not present on the wizard

![area](https://user-images.githubusercontent.com/3728712/39314895-fdcfd402-497d-11e8-90be-df4458187c4a.png)

PR #465, opened 2018-04-27
HtmlArea content with CKE - incorrect behavior of validation, when re…

ashklianko:
…quired inputs in wizard #461


ashklianko:
@sgauruseu  please check

Issue #492, opened 2018-05-10
Replace the menu icon in the Mobile preview header


pmi:
@alansemenov commented on [Tue May 08 2018](https://github.com/enonic/app-contentstudio/issues/95)

* Change height of the header from 44px to 36px
* Vertically center display name and all the buttons
* Remove the status line (status will be in the toolbar underneath, see https://github.com/enonic/app-contentstudio/issues/82)
* Replace the dropdown icon to the right of the title with the new icon as shown below:
![image](https://user-images.githubusercontent.com/11612490/39752549-1881d114-52bc-11e8-8799-8f177beacf92.png)
* When the menu is expanded, flip the new icon:
![image](https://user-images.githubusercontent.com/11612490/39753189-33896bd2-52be-11e8-8152-e21fcf048d02.png)

PR #493, opened 2018-05-10
Replace the menu icon in the Mobile preview header #492

pmi:


pmi:
Moved icons to app-contentstudio so this is no longer needed

Issue #479, opened 2018-05-04
Implement new dropdown control for sorting dialog


alansemenov:
The list of options in this dropdown is getting long, we duplicate every option except for one and not everyone understands what "Ascending" and "Descending" is.

What we need is one option per sorting alternative inside the dropdown, remove "Ascending"/"Descending" from option titles and for each option (except for "Manually sorted"!) show two icons representing sorting order (to the left of the option title). The icons should be absolutely positioned so that they are correctly vertically aligned against all options.

We already have icons representing sorting order for alpha and numeric fields that we are using in the Content grid - we can try reusing those and see how it looks.

Indicate current sorting option/order by highlighting correct icon (use the blue color we are using in toggle icons). Use has to explicitly click on an icon in order to change sorting. Clicking on the option title or an already highlighted icon shouldn't do anything. Make sure the icons are big enough and there's enough padding around them so that they are easily clickable even on mobile.

Show "Sort in ascending order" and "Sort in descending order" as tooltips for the icons.

![image](https://user-images.githubusercontent.com/11612490/39628558-c013b2e6-4fa9-11e8-89da-b393484ebf2d.png)

PR #499, opened 2018-05-14
Implement new dropdown control for sorting dialog #479

edloidas:
Added customization of the TabMenu.

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/tab/TabItem.ts
+++ b/src/main/resources/assets/admin/common/js/ui/tab/TabItem.ts
@@ -41,7 +41,9 @@ module api.ui.tab {
                 this.setFocusable(false);
             }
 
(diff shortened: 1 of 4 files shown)


alansemenov:
Great stuff

sigdestad:
Only thing is that the arrows should point up and down (now both point down)

sigdestad:
Or, down and up it should be I guess :-)

edloidas:
Well, it's either we use these icoMoon icons or use the arrows only. You see, they were designed with changes the order of the numbers (`1..9` => `9..1`) or letters (`A..Z` => `Z..A`). We don't have any other variations. BTW we use those icons for a long time now in the grid to show the content order.

alansemenov:
Yes, the icons were designed this way for a reason. Arrows represent the direction the items are listed in.

Issue #494, opened 2018-05-11
CK editor must inherit and use same fonts as the rest of content studio


ashklianko:
@sigdestad :                       
CK editor must inherit and use same fonts as the rest of content studio as default (currently looks like Times roman or similar). See how this was done in Tiny as styling is correct there.

PR #502, opened 2018-05-14
CK editor must inherit and use same fonts as the rest of content stud…

ashklianko:
…io #494

-aligned input type's cke styling with tiny


ashklianko:
Aligned input types styling to be same as for tiny; live edit seems ok

PR #613, opened 2018-06-15
Issue details dialog - users without publish rights should be removed…

pmi:
… from the Assignees selector #612


sigdestad:
Ups? What is this task? There should definetly not be a requirement to be a publisher?

pmi:
@sigdestad Nope, that's not about publishing rights, that's about rights to access content studio app. Because it looks strange that you are allowed to choose these users, but they are not saved eventually because of the backend check.

sigdestad:
Great :-)

PR #617, opened 2018-06-18
CKEditor: Improve UX on drag-n-drop uploading of images by giving vis…

ashklianko:
…ual queues when dragging #616

-Showing border around editor when dragging over it (only in wizard, not working in inline mode)

### DIFF
--- a/src/main/resources/assets/admin/common/js/util/htmlarea/editor/HTMLAreaBuilderCKE.ts
+++ b/src/main/resources/assets/admin/common/js/util/htmlarea/editor/HTMLAreaBuilderCKE.ts
@@ -176,6 +176,7 @@ module api.util.htmlarea.editor {
             this.setupKeyboardShortcuts(ckeditor);
             this.addCustomLangEntries(ckeditor);
             this.removeUnwantedMenuItems(ckeditor);
+            this.updateEditorDragOverStyling(ckeditor);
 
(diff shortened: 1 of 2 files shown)


ashklianko:
Hi @Bellfalasch and @alansemenov !
Added some dashed border around editor:
image

Tell me if that is good enough!

It works only in wizard right now, in case it is also needed in inline mode I'd create a separate task because some moments to be discussed.

alansemenov:
Hmm, maybe you misunderstood. I believe what @Bellfalasch asked for in the first place was the same pulsating blue circle that we get when you drag over an image over image selector.

screen shot 2018-06-18 at 15 21 21

I guess the main challenge here is to position this circle properly, so that it doesn't stand in the way when the user is moving the cursor. Ideally, it should appear when there's a delay in DnD - user is not moving the mouse but still holds the button - then we could show the circle, but hide it whenever user starts moving the cursor again.

Bellfalasch:
I was after the pulsating blue "drop indicator" yes. But if it is very problematic perhaps it should go via PAB instead and get a proper spec. This was just a tiny idea on UX-improvement that popped up when testing the new editor.

ashklianko:
Yes, I understood this, just possibly option that i propose would be more convenient for user rather then blue circle following cursor, at least user will see that he can drag?

PR #625, opened 2018-06-20
Fetch dragged URLs from the ImageSelector #606

ashklianko:
-Extracting img src from dragged data, requesting img and sending it to uploader.
Thanks to ckeditor's uploadimage plugin


ashklianko:
That's what we actually do :)

alansemenov:
Ok, I see. It seems to work but it's very buggy. After I dragged the first image it got uploaded, but there were tons of 404 errors in the console (see below). Looks like some assets are missing? But it was created as a content and it does show in the editor and everything looks ok. However, when I dragged the second image below the first one, it was also created as a content but it didn't show in the editor. I have to remove the first image, and then I can drag another one and it will show.

image

Issue #674, opened 2018-07-04
Date/time picker dropdown is cut off with no scroller


edloidas:
@sgauruseu commented on [Wed Jul 04 2018](https://github.com/enonic/app-contentstudio/issues/289)

1. Select a site and click on `Schedule...`

**Issue** - Ok button for `Online to` is not available

![image](https://user-images.githubusercontent.com/3728712/42268587-91c34c8e-7f84-11e8-85de-f9c4e8cde48f.png)

---

@alansemenov commented on [Wed Jul 04 2018](https://github.com/enonic/app-contentstudio/issues/289#issuecomment-402427929)

The button is there but the dropdown is cut off due to screen height and there's no scrollbar. On mobile it's even worse (see below). We should do the same thing we do for regular dropdowns - calculate space below and above and see where the div will fit. If neither space is available, open underneath but with a scrollbar. 

![image](https://user-images.githubusercontent.com/11612490/42270126-5d91d0c4-7f80-11e8-8c74-3b0ea012a0ec.png)

PR #675, opened 2018-07-04
Scheduled Publishing dialog - button 'Ok' is not available, when the …

edloidas:
…dialog is opened on laptop #674

Added position and height calculations for the picker dialogs.
Updated version of the `enonic-admin-artifacts`, fixing `picker` mixin.


alansemenov:
Nice!

PR #685, opened 2018-07-06
some cke plugins are missing in build

ashklianko:


ashklianko:
@sgauruseu found that some dialogs do not open when server is not in dev mode

PR #712, opened 2018-07-17
Issue 631

ashklianko:


src/main/resources/assets/admin/common/js/util/htmlarea/editor/HTMLAreaBuilderCKE.ts
```diff
@@ -486,12 +486,14 @@ module api.util.htmlarea.editor {
...
             ckeditor.addCommand('openMacroDialog', {
-                exec: (editor) => {
-                    this.notifyMacroDialog(editor);
+                exec: (editor, data: any) => {
+                    this.notifyMacroDialog({editor: editor, macro: data});
                     return true;
                 }
             });
 
+            CKEDITOR.plugins.addExternal('macro', this.assetsUri + '/admin/common/js/util/htmlarea/plugins/', 'macroCKE.js');
```
alansemenov: We should not hardcode "_/admin_" anywhere in paths. Use ``CONFIG.adminUrl`` for that. Other than that, everything is fine and works like a charm.

ashklianko: Will update

alansemenov: Oh, sorry - this is not "_/admin_" that is url of the admin console, this is actual path of the asset path. Everything is fine here.

PR #722, opened 2018-07-27
Extract references from HTML Area by parsing its contents #721

vbradnitski:


src/main/resources/assets/admin/common/js/form/inputtype/text/HtmlAreaCK.ts
```diff
@@ -404,9 +405,37 @@ module api.form.inputtype.text {
...
+            while (imageMatch) {
+                const imageReferenceId = imageMatch.length == 4 ? imageMatch[3] : null;
+
+                if (imageReferenceId) {
+                    valueObj.getPropertySet().addReference('references', Reference.from(new api.content.ContentId(imageReferenceId)));
+                }
+                imageMatch = imageRegexp.exec(value);
+            }
+        }
+
```
alansemenov: We should refactor this. The only difference between extracting image references and other content references is regex that we are using for parsing.
So, make a function called "extractReferences" which will take a regex as an argument and call this function twice from parseReferences, first time passing contentRegexp and second time passing imageRegexp.

Issue #716, opened 2018-07-26
Toolbar button is not highlighted when image is selected in HTML Area


alansemenov:
1. Insert an image into HTML Area
2. Single-click the image inside the editor

The toolbar icon for the image dialog is supposed to be highlighted (same as for macro) but it's not.

PR #726, opened 2018-07-30
Toolbar button is not highlighted when image is selected in HTML Area…

ashklianko:
… #716


ashklianko:
WIll fix all toolbar buttons together in one place

PR #735, opened 2018-08-01
Fetch dragged URLs from the ImageSelector #606

ashklianko:
-Extracting img data src from dragged data, requesting img and sending it to uploader.
Thanks to ckeditor's uploadimage plugin


ashklianko:
Working on cross origin images dragging. Currently local files dragged work and images with src encoded to base64 like in google search

ashklianko:
Also see https://github.com/enonic/xp/pull/6488

Issue #768, opened 2018-08-21
Issue Detail Dialog - Tooltip stays  when the dialog has been escaped


ashklianko:
@sgauruseu commented on [Wed Jul 11 2018](https://github.com/enonic/app-contentstudio/issues/314)

1. Open Issue Details Dialog , move cursor to the Comment link
Tooltip appears(OK)
2. Press `Esc` key twice and close modal dialogs
Bug - modal dialog has been closed, but tooltip is still visible 

![image](https://user-images.githubusercontent.com/3728712/42577083-76755b7c-852c-11e8-98f0-e2ee78e1324f.png)

PR #772, opened 2018-08-22
Issue Detail Dialog - Tooltip stays when the dialog has been escaped …

ashklianko:
…#768

-When elements are removed from DOM and then added back they're not reregistered in ElementRegistry; tooltip handler uses ElementRegistry to find elements and bind removed/hidden events to them, thus elements not found


ashklianko:
https://github.com/enonic/app-contentstudio/pull/381 is required

PR #864, opened 2018-11-05
Support preselect in BucketView

ashklianko:
…#863

Changes reqired for https://github.com/enonic/app-contentstudio/issues/490 to work

### DIFF
--- a/src/main/resources/assets/admin/common/js/aggregation/BucketAggregationView.ts
+++ b/src/main/resources/assets/admin/common/js/aggregation/BucketAggregationView.ts
@@ -43,6 +43,17 @@ module api.aggregation {
             return isSelected;
         }
 
+        selectBucketViewByKey(key: string, supressEvent?: boolean) {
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/aggregation/BucketAggregationView.ts
```diff
@@ -43,6 +43,17 @@ module api.aggregation {
...
+            this.bucketViews.some((bucketView: api.aggregation.BucketView) => {
+                if (bucketView.getName() === key) {
+                    bucketView.select(supressEvent);
+                    return true;
+                }
+
+                return false;
+            });
+        }
+
```
alansemenov:
```
if (this.bucketViews.some((bucketView: api.aggregation.BucketView) => bucketView.getName() === key)) {
  bucketView.select(supressEvent);
}

```

alansemenov: Ah no, this won't work - ``bucketView`` is outside of the function scope.

PR #892, opened 2019-01-08
"Show page editor" button won't open Page Editor after details panel …

ashklianko:
…is expanded in the Wizard #891

-need to fetch splitter thickness


ashklianko:
Required for https://github.com/enonic/app-contentstudio/issues/511

Issue #895, opened 2019-01-14
Confirm deletion of a changed item/option-set occurrence


alansemenov:
As of now, it's very easy to delete an occurrence of an item-set or option-set simply by clicking the "x" icon. Sometimes users have very complex structure of items inside their sets and we should ask for confirmation whenever user tries to delete a "dirty" item/option set.

* Implement a _dirty_ flag (``=false`` by default) inside ``api.form.FormSetOccurrenceView``
* Set it to ``true`` whenever any input inside the set has been changed (we could've checked if PropertySet is empty, but inputs might have default values and the set with default values shouldn't be considered "dirty")
* Intercept an attempt to delete an occurrence, check if the set is "dirty" and, if it is, ask user to confirm deletion: "_The form data has been changed. Are you sure you want to delete it?_" (remember to localise the resource). "Yes" button will delete the occurrence, "No" will cancel deletion.

PR #898, opened 2019-01-18
Confirm deletion of a changed item/option-set occurrence #895

ashklianko:
-Listening PropertyValueChangedEvent to know when any FormItemView data in FormSetOccurrenceView was updated
-Maintaining map like object to keep modified FormItemViews with their original and current value (FormItemView is added to map only after PropertyValueChangedEvent); When value reset back to original - removing FormItemView from map object
-If delete occurrence button pressed then checking if map object is empty or not

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
+++ b/src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
@@ -2,6 +2,9 @@ module api.form {
 
     import PropertySet = api.data.PropertySet;
     import PropertyArray = api.data.PropertyArray;
+    import PropertyValueChangedEvent = api.data.PropertyValueChangedEvent;
+    import ConfirmationDialog = api.ui.dialog.ConfirmationDialog;
+    import i18n = api.util.i18n;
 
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
```diff
@@ -23,20 +26,63 @@ module api.form {
...
+                .setQuestion(i18n('dialog.confirm.occurrences.delete'))
+                .setYesCallback(() => {
+                    this.notifyRemoveButtonClicked();
+                });
+        }
+
+        private initFormDataChangeListener() {
+            this.formDataChangedListener = (event: PropertyValueChangedEvent) => {
+                this.formItemViews.filter(
+                    (formItemView: FormItemView) => formItemView.getFormItem().getName() === event.getProperty().getName()).forEach(
```
alansemenov: Move ``event.getProperty().getName()`` and ``event.getNewValue()`` up, out of ``filter``/``forEach`` loops and assign to const - no point in fetching these values in each iteration. Same with ``formItemView.getFormItem().getName()`` - it's called 4 times inside the ``forEach`` loop.

Issue #903, opened 2019-01-23
Ask for confirmation when deleting existing item/option-set


alansemenov:
1. Adjustment to new functional of #895 

In addition to what we do now, ask for confirmation when deleting an existing (saved) set with any (dirty or not) values inside. 

Test cases:
* A new set with no values: no confirmation
* A new set with no dirty fields (ie only default values): no confirmation
* A new set with dirty fields: confirmation
* A saved set with no dirty fields and no values: no confirmation
* A saved set with no dirty fields but with values: confirmation
* A saved set with dirty fields: confirmation

2. Changes to the confirmation dialog
Title: "_Remove ``<item-set.label>?``_"
Body: _Are you sure you want to remove this item?_

PR #915, opened 2019-01-29
Ask for confirmation when deleting existing item/option-set #903

ashklianko:
-Reworked FormSetOccurrenceView to correctly check it's properties are dirty
-added hasNonDefaultValues() method in item view and occurrences view classes to check if there are non empty/non default values in form items to trigger confirmation dialog in case when occurrence removed


alansemenov:
Great stuff

PR #966, opened 2019-03-23
Correctly handle empty property set

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/966

PR #1000, opened 2019-05-16
Move PropertyTreeHelper from CS to lib-admin-ui

ashklianko:
Moving PropertyTreeHelper.ts from contentstudio to lib-admin-ui thus any app could use it.


ashklianko:
Needed for:
https://github.com/enonic/app-users/pull/235
https://github.com/enonic/app-contentstudio/pull/481

Issue #1004, opened 2019-05-28
Notify about data refresh after fetching tree nodes


ashklianko:
@sgauruseu commented on [Tue May 14 2019](https://github.com/enonic/app-applications/issues/94)

1. Select 2 started applications in the grid
2. Press the `Stop` button

Bug: 
**Expected** - Start button gets enabled and Stop button gets disabled
**Actual** -    Start button is disabled and Stop button is enabled. Incorrect state is on the  `ApplicationItemStatisticsPanel`

![image](https://user-images.githubusercontent.com/3728712/57698184-b0b1f700-765d-11e9-993d-46c728958f68.png)

ashklianko:
Changes required in treegrid to complete this task

PR #1005, opened 2019-05-28
Notify about data refresh after fetching tree nodes

ashklianko:
…r starting 2 selected applications #1004

Need to update treegrid to complete https://github.com/enonic/app-applications/issues/94


ashklianko:
Required for https://github.com/enonic/app-applications/issues/94

PR #1021, opened 2019-06-21
Implement a method to sync empty arrays in a property set

vbradnitski:
…rd #1020

### DIFF
--- a/src/main/resources/assets/admin/common/js/data/PropertySet.ts
+++ b/src/main/resources/assets/admin/common/js/data/PropertySet.ts
@@ -473,20 +473,16 @@ module api.data {
                     this.addPropertyArray(propertyArray.copy(this));
                 }
 
+                const thisPropertyArray = this.getPropertyArray(propertyArray.getName());
+
                 propertyArray.forEach(((property, index) => {
                     if (property.getType().equals(api.data.ValueTypes.DATA)) {
 
-                        const targetPropertySet = property.getValue().getPropertySet();
+                        const targetPropertySet = property.getPropertySet();
+                        const thisPropertySet = thisPropertyArray.getSet(index);
 
-                        const thisPropertyArray = this.getPropertyArray(propertyArray.getName());
-                        const thisProperty = thisPropertyArray.get(index);
-                        if (thisProperty && thisProperty.getType().equals(api.data.ValueTypes.DATA)) {
-
-                            const thisPropertySet = thisProperty.getValue().getPropertySet();
-
-                            if (thisPropertySet) {
-                                thisPropertySet.syncEmptyArrays(targetPropertySet);
-                            }
+                        if (thisPropertySet) {
+                            thisPropertySet.syncEmptyArrays(targetPropertySet);
                         }
                     }
                 }));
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/admin/common/js/data/PropertySet.ts
```diff
@@ -461,6 +461,38 @@ module api.data {
...
+                    if (property.getType().equals(api.data.ValueTypes.DATA)) {
+
+                        const targetPropertySet = property.getValue().getPropertySet();
+
+                        const thisPropertyArray = this.getPropertyArray(propertyArray.getName());
+                        const thisProperty = thisPropertyArray.get(index);
+                        if (thisProperty && thisProperty.getType().equals(api.data.ValueTypes.DATA)) {
+
+                            const thisPropertySet = thisProperty.getValue().getPropertySet();
+
```
alansemenov: same here - ``thisProperty.getPropertySet()``

Actually, instead of doing 
```
const thisProperty = thisPropertyArray.get(index);
const thisPropertySet = thisProperty.getValue().getPropertySet();
```
you could simply do
```
const thisPropertySet = thisPropertyArray.getSet(index);
```

src/main/resources/assets/admin/common/js/data/PropertySet.ts
```diff
@@ -461,6 +461,38 @@ module api.data {
...
         }
 
+        public syncEmptyArrays(target: PropertySet) {
+
+            if(!target) {
+                return;
+            }
+
+            target.getPropertyArrays().forEach((propertyArray: PropertyArray) => {
+                if (!this.getPropertyArray(propertyArray.getName())) {
```
alansemenov: here we can store value of the array:
```
const thisPropertyArray = this.getPropertyArray(propertyArray.getName());
if (!thisPropertyArray) {
  this.addPropertyArray(propertyArray.copy(this));
  thisPropertyArray = this.getPropertyArray(propertyArray.getName());
}

File: src/main/resources/assets/admin/common/js/data/PropertySet.ts [hunk 3]
```diff
@@ -461,6 +461,38 @@ module api.data {
...
+
+                    this.addPropertyArray(propertyArray.copy(this));
+                }
+
+                propertyArray.forEach(((property, index) => {
+                    if (property.getType().equals(api.data.ValueTypes.DATA)) {
+
+                        const targetPropertySet = property.getValue().getPropertySet();
+
```
alansemenov: ``PropertyValue`` class has ``getPropertySet()`` method, so you can do ``property.getPropertySet()`` directly.

src/main/resources/assets/admin/common/js/data/PropertySet.ts
```diff
@@ -461,6 +461,38 @@ module api.data {
...
+                        const thisPropertyArray = this.getPropertyArray(propertyArray.getName());
```
alansemenov: ...and then use ``thisPropertyArray`` here instead of declaring it every time

PR #1116, opened 2019-10-09
Add possibility to not close modal dialog on click outside of dialog …

ashklianko:
…#1115


sigdestad:
NOTE: We still want to close the fulltext mode if a user clicks outside of the modal (as usual).

PR #1156, opened 2019-11-22
Fix handling of child dialogs spawned by a modal dialog

ashklianko:
Issue origin: dialog can be spawned within other parent dialog without notifying it (e.g. in SiteConfiguratorDialog), so when handlers that check if user clicked out of a parent dialog invoke `hasSubDialog()` it returns false meanwhile other dialog is open, thus focus starts jumping between these 2 open dialogs.
Solved by updating modal dialog and keeping trace of open dialogs independently:
1. Added class with static methods (DialogManager) that will hold list of open dialogs and will notify on new dialog opened. Any class can use it to listen for dialog open event or check number ot total open dialogs
2. Modal dialog will update DialogManager on opening and closing
3. Added modal dialog state enum (OPEN|CLOSED) so now you could use `isOpen()` method instead of `isVisible()` for dialog. Also not invoking clickOutside callback in case when dialog is masked

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/dialog/ModalDialog.ts
+++ b/src/main/resources/assets/admin/common/js/ui/dialog/ModalDialog.ts
@@ -816,6 +816,8 @@ module api.ui.dialog {
 
         private openDialogs: ModalDialog[];
 
+        private maskedBy: Map<string, ModalDialog[]>;
+
         private dialogOpenListeners: { (dialog: ModalDialog): void; } [];
 
         private static INSTANCE: DialogManagerInner;
@@ -823,6 +825,7 @@ module api.ui.dialog {
         private constructor() {
             this.openDialogs = [];
             this.dialogOpenListeners = [];
+            this.maskedBy = new Map();
         }
 
         public static get(): DialogManagerInner {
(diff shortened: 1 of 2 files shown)


ashklianko:
UPDATED: 

-Revised usage of mask/unmask method: now DialogManager will control masking/unmasking of dialogs itself; e.g. if you open dialog while other dialog is open then other dialog will be automatically masked; when dialog is closed then DialogManager will unmask dialogs that were masked on opening this dialog
-Now DialogManager is not used externally but will remain for future use; mask/unmask methods also left public for possible future use

src/main/resources/assets/admin/common/js/ui/dialog/ModalDialog.ts
```diff
@@ -806,4 +812,108 @@ module api.ui.dialog {
...
+
+        public static get(): DialogManagerInner {
+            if (!DialogManagerInner.INSTANCE) {
+                DialogManagerInner.INSTANCE = new DialogManagerInner();
+            }
+
+            return DialogManagerInner.INSTANCE;
+        }
+
+        opened(dialog: ModalDialog) {
```
alansemenov: `opened` and `closed` are not good names for these methods. `handleOpenDialog` and `handleClosedDialog` should be fine.

Issue #1229, opened 2020-03-27
Refactor usage of save confirmation dialog


alansemenov:
We now have ConfirmationDialog in `lib-admin-ui` used by SettingsDataItemWizardPanel to show a confirmation about unsaved changes in the `app-contentstudio`'s Project Settings wizard, that looks like this:

![image](https://user-images.githubusercontent.com/11612490/77752115-3eed5c80-7027-11ea-9418-a569c09bcb89.png)

and sometimes (when the project is invalid) like this:

![image](https://user-images.githubusercontent.com/11612490/77751877-c5ee0500-7026-11ea-9ed6-94e143c96f93.png)

This dialog is using localisation phrases from both `lib-admin-ui` (`dialog.saveBeforeClose.msg`) and `app-contentstudio` (`dialog.confirmClose.invalidData`).

We also have SaveBeforeCloseDialog in app-users, doing exactly the same thing as the first one, looking like this:

![image](https://user-images.githubusercontent.com/11612490/77752153-53c9f000-7027-11ea-82f3-3ec7f1beec4b.png)

This dialog is using local phrases `dialog.saveBeforeClose.title` and `dialog.saveBeforeClose.msg`, one of which is a duplicate of the phrase with the same name from lib-admin-ui.

This is very chaotic.

What we need is:
1. Use Confirmation dialog everywhere (dialog body text should still be set via `.setQuestion()`).
2. In Confirmation add support for keyboard shortcuts to click the buttons (like it's done in SaveBeforeCloseDialog)
3. Use only two common phrases in `lib-admin-ui`: `dialog.confirm.title` and `dialog.confirm.unsavedChanges` (rename `dialog.saveBeforeClose.msg`). There's no reason to have two different texts for when data is valid or invalid.
4. Remove all related phrases from `app-users` and `app-contentstudio`
5. Switch app-users to use Confirmation dialog and remove SaveBeforeCloseDialog
6. Make sure styles of the confirmation dialog are only in lib-admin-ui, remove all related styling from the other two apps.

PR #1230, opened 2020-03-30
Refactor usage of save confirmation dialog #1229

ashklianko:
-NB: Had to move addAction() calls into initElements(), otherwise if triggered after ModalDialog's open() key bindings won't be set

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/dialog/ConfirmationDialog.ts
+++ b/src/main/resources/assets/admin/common/js/ui/dialog/ConfirmationDialog.ts
@@ -27,8 +27,6 @@ export class ConfirmationDialog
     doRender(): Q.Promise<boolean> {
         return super.doRender().then((rendered) => {
             this.appendChildToContentPanel(this.questionEl);
-            this.addAction(this.yesAction, true);
-            this.addAction(this.noAction);
 
             return rendered;
         });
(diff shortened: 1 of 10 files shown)


src/main/resources/assets/admin/common/js/ui/dialog/ConfirmationDialog.ts
```diff
@@ -67,8 +65,15 @@ export class ConfirmationDialog
...
 
         this.questionEl = new H6El('question');
-        this.noAction = new Action(i18n('action.no'), 'esc');
-        this.yesAction = new Action(i18n('action.yes'));
+
+        this.noAction = new Action(i18n('action.no'), i18n('action.no').slice(0, 1).toLowerCase());
+        this.noAction.setMnemonic(i18n('action.no').slice(0, 1).toLowerCase());
+
+        this.yesAction = new Action(i18n('action.yes'), i18n('action.yes').slice(0, 1).toLowerCase());
+        this.yesAction.setMnemonic(i18n('action.yes').slice(0, 1).toLowerCase());
```
alansemenov: Can we avoid calling `i18n` twice for the same key? This will actually browse all resources every time it's called...

ashklianko: done

PR #1275, opened 2020-05-08
Browse toolbar is not updated after deleting filtered content #1273

ashklianko:
-Keeping list of selected data items ids, NOT node ids. Thus no need to keep list of current and stashed selection and filter duplicated nodes when getting selected nodes
-BrowsePanel: treeNodesToBrowseItems -> dataItemsToBrowseItems since only node data was used in browse panels
-Left 1 function to deselect all rows instead of 3!
-RootNode and TreeNode is internal grid structure thing, should not be exposed to non-inheritors; Updated so non-inheritors get data items directly instead of nodes (most of them just needed to count items). Still a lot to be done


ashklianko:
To be done:
-all treeroot access from non treegrid inheritors to be removed
-refactor onhighlightingChanged
-all interaction with treegrid via treegrid api, applying Demeter's Law

PR #1311, opened 2020-05-28
[IE11] Action menu in mobile preview doesn't close

alansemenov:


ashklianko:
By some reason actions in fold menu are not executed. Checking...

PR #1483, opened 2020-09-30
MenuButton - add method enable automatic menu toggling on main action…

pmi:
… #1482

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
+++ b/src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
@@ -65,8 +65,8 @@ export class MenuButton
         this.menu.removeSeparator();
     }
 
-    toggleMenu(expanded?: boolean) {
-        if (expanded || !this.menu.hasClass('expanded')) {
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
```diff
@@ -63,6 +65,20 @@ export class MenuButton
...
+        } else {
+            this.collapseMenu();
+        }
+    }
+
+    expandMenu(): void {
+        this.menu.toggleClass('expanded');
+        this.dropdownHandle.toggleClass('down');
+        this.dropdownHandle.giveFocus();
+    }
```
alansemenov, line 80: The method is called "`expandMenu`", but its code actually toggles the menu.

`expandMenu` and `collapseMenu` should have similar code with reverse logic. Either `expandMenu` should use `addClass` (like `collapseMenu` is using `removeClass` now), or both of them should use `toggleClass`, but with different second argument.

alansemenov (changes requested):

alansemenov (approved):

PR #1568, opened 2020-11-11
Dynamically set title of an option-set based on selected options #1567

pmi:


src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
```diff
@@ -62,25 +63,34 @@ export class FormItemSetOccurrenceView
...
-                this.label.setText((<HTMLInputElement>firstNonEmptyInput).value);
-            } else {
-                this.label.setText(StringHelper.htmlToString(firstNonEmptyInput['value']));
-            }
-            this.formSetOccurrencesContainer.unDescendantAdded();
+        if (selectedValues.length === 0) {
+            return this.getFormSet().getLabel();
         } else {
-            this.label.setText(this.getFormSet().getLabel());
+            return selectedValues[0];
```
alansemenov, line 80: `return selectedValues[0] || this.getFormSet().getLabel();`

src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
```diff
@@ -62,25 +63,34 @@ export class FormItemSetOccurrenceView
...
         }
     }
+
+    private recursiveFetchLabels(propArray: PropertyArray, labels: string[], firstOnly?: boolean): void {
+        propArray.forEach((prop: Property) => {
```
alansemenov, line 85: We don't really need the `firstOnly` argument because it's never used (it's always true).

src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
```diff
@@ -62,25 +63,34 @@ export class FormItemSetOccurrenceView
...
+            if (ValueTypes.STRING.equals(prop.getType()) && prop.getValue().isNotNull()) {
+                labels.push(prop.getString());
+                if (firstOnly) {
+                    return;
+                }
```
alansemenov, line 90: `return;`

src/main/resources/assets/admin/common/js/form/set/optionset/FormOptionSetOccurrenceView.ts
```diff
@@ -121,6 +133,34 @@ export class FormOptionSetOccurrenceView
...
+            });
+        }
+
+        if (selectedLabels.length === 0) {
+            return this.getFormSet().getLabel();
+        } else {
+            return selectedLabels.join(', ');
+        }
+    }
+
```
alansemenov, line 163:
```
    protected getLabelText(): string {
        const selectionArray = this.propertySet.getPropertyArray('_selected');
        let selectedLabels: string[] = [];

        if (selectionArray && !selectionArray.isEmpty()) {
            const nameLabelMap = new Map(
                this.getFormItems()
                    .filter((formItem: FormItem) => formItem instanceof FormOptionSetOption)
                    .map((formItem: FormOptionSetOption) => [formItem.getName(), formItem.getLabel()] as [string, string])
            );

            selectedLabels = selectionArray.map((selectedProp: Property) => nameLabelMap.get(selectedProp.getString()));
        }

        return selectedLabels.length ? selectedLabels.join(', ') : this.getFormSet().getLabel();
    }
```

alansemenov (changes requested):

alansemenov:
Nice job! See a few comments above. Also, I noticed that for Multi-select option-set the order of options labels inside the title is based on the order you selected the options. Can this be fixed so that the order is based on the order of options inside the set?

image

src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
pmi: It's for possible future use, in case we ever need to get more labels nested inside.

src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
pmi: yes, because the method returns void and we want to quit.

src/main/resources/assets/admin/common/js/form/set/itemset/FormItemSetOccurrenceView.ts
pmi: Hm, thought that will produce array out of bounds exception, but it seems to work.

pmi: Don't like it still :)

pmi:
Added sorting based on original order

alansemenov (approved):

Issue #1617, opened 2020-12-02
Change confirmation for deleting an occurrence


alansemenov:
We now show this confirmation dialog before deleting an occurrence:
![image](https://user-images.githubusercontent.com/11612490/100890965-1290b680-34b9-11eb-940d-b3148e55c3f6.png)

Remove this (it won't work with current custom label anyway) and do the following instead:
1. Expand the `form-option-set-occurrence-view` element, unless it's already expanded.
2. Apply a mask to the entire `form-option-set-occurrence-view` element
3. Show two buttons on top of the mask (align them with the top of the `form-option-set-occurrences-container` element): _Delete_ and _Cancel_. No need to display the label inside the _Delete_ button.

When _Delete_ is clicked, delete the occurrence. 

If the occurrence form is empty, the occurrence will be instantly deleted without confirmation, just like now.

![image](https://user-images.githubusercontent.com/11612490/100891246-5daac980-34b9-11eb-9a96-bb51fac1fc95.png)

PR #1637, opened 2020-12-10
Change confirmation for deleting an occurrence #1617

pmi:


alansemenov (changes requested):
Two things to fix:

1. Use option-set's label on the "Delete ..." button, since we don't have that much space there to use labels of selected options. Also, make sure to set some sensible max-width and ellipsis overflow for this button, so that it doesn't get too wide.
image

2. Clicking "Collapse all" link while the confirmation dialog is visible will collapse the sets but the mask will remain.

pmi:
Done !

alansemenov (approved):

Issue #1679, opened 2021-01-04
No content validation after changing option in a single-select option-set


sgauruseu:
1. Select Features and open new Option Set wizard.
2. Fill in the name input and select `option 2` in Single Selection, save all.
![image](https://user-images.githubusercontent.com/3728712/103511011-46277d80-4e77-11eb-8b67-aa9d7d91f58a.png)

3. Select `Option 1` in Single Selection

**BUG**: 
- `Saved` button remains disabled in the toolbar after updating the option.
-  Red icon should appear in the wizard, because the content gets not valid after the updating
![image](https://user-images.githubusercontent.com/3728712/103511258-be8e3e80-4e77-11eb-9a00-fc33e156ff07.png)

**Expected behaviour :**  Save is enabled and the content is not valid

![image](https://user-images.githubusercontent.com/3728712/103511359-f4332780-4e77-11eb-9afb-29dd08f02dfd.png)

PR #1686, opened 2021-01-05
No validation after a different option was selected in an option-set …

pmi:
…#1679


alansemenov (changes requested):
This works well, but only partially (and the original description of the task is also not entirely correct @sgauruseu).

Yes, when user changes an option (like anything else in the content wizard), Save button should become enabled.
Yes, when user switches to an option that has an invalid form (empty required fields) then the content should be marked as invalid.

But no, we should not run validation of an **empty** option form before user even gets a chance to fill out the fields.

Seeing this right after you selected an option is not a good user experience:
image

It should be like with other inputs: invalid option-set should be highlighted after invalid content is saved or when invalid content is opened.

PR #1694, opened 2021-01-07
Option Set wizard - Esc won't close the delete confirmation mask #1693

pmi:


alansemenov (changes requested):
The problem is wider than originally described. As of now, the mask won't intercept Tab commands. So if you start tabbing when the mask opens, you'll see that tab will navigate through elements behind the mask, and that is wrong. 

Also, doing anything on Enter doesn't make much sense now since user doesn't see which of the two buttons has a focus, so it's not obvious what Enter will do. 

What we should do is this when the mask opens is this:
1. Set focus on the confirm button once the mask opens, so that it looks like below
2. Make sure that Tab toggles focus between the two buttons and nothing else 
3. Intercept Enter and execute action on the focused button
4. Intercept Esc and close the mask (this seems to work now)

image

alansemenov (approved):

Issue #1684, opened 2021-01-05
Fix Collapse links in Option-set


alansemenov:
* We don't need Collapse/Expand/Collapse all/Expand all links when there's just one occurrence (minimum=1 & maximum=1)
* We shouldn't show Collapse link for an occurrence when option is not selected
* We shouldn't show Collapse link for an occurrence when option is selected but the form is empty

Case 1: "Collapse" is shown when no option is selected

![image](https://user-images.githubusercontent.com/11612490/103654422-228f3080-4f66-11eb-8157-86b378f79c67.png)

Case 2: "Collapse all" links are shown when no option is selected in any of the occurrences

![image](https://user-images.githubusercontent.com/11612490/103654490-3b97e180-4f66-11eb-929f-fd4e3fdede7e.png)

Case 3: "Collapse" link is shown when option is selected but there's no form

![image](https://user-images.githubusercontent.com/11612490/103654576-5b2f0a00-4f66-11eb-91b9-98040f70deac.png)

When there's more than one occurrence, there's also some extra space at the bottom:

![image](https://user-images.githubusercontent.com/11612490/103654690-7dc12300-4f66-11eb-809d-07c02d88fabb.png)

PR #1697, opened 2021-01-08
Fix Collapse links in Option-set #1684

pmi:


alansemenov (changes requested):
Much better than before but here's a bug:
1. Change the single-select option-set so that it allows multiple occurrences
2. Select "Option 2" (which doesn't have a form).
3. Add second occurrence.
4. Select "Option 1". Its form expands right away.
5. The links appear saying "Expand all", even though first occurrence doesn't have a form and the second one has its form expanded:

image

alansemenov (changes requested):
Not 100% there yet :)
The case above is fixed now, but if you repeat the same steps and then add a THIRD occurrence, then you get exactly the same problem.

Here everything is ok, except for the extra empty space at the bottom which is still not fixed. `form-option-set-occurrences-container` has to be hidden until an option is selected.
image

But here is how it looks after I selected "Option 1" in the 2nd occurrence, then clicked "Collapse all" and then added a 3rd occurrence:
image

Exactly the same problem: the first two occurrences are collapsed and the new one is still empty, yet the links say "Collapse all" even though there's nothing to collapse.

It looks like a new occurrence is treated like expanded **before option is selected**, and that's why we get extra-space at the bottom and inverted behaviour of the links.

Please test this thoroughly before committing.

src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
```diff
@@ -225,13 +229,19 @@ export abstract class FormSetOccurrenceView
...
         this.toggleClass('collapsed', !show);
         this.label.setTitle(i18n(show ? 'tooltip.header.collapse' : 'tooltip.header.expand'));
     }
 
     isContainerVisible(): boolean {
-        return this.formSetOccurrencesContainer.isVisible();
+        // container may be on, but will be not visible
+        // if the whole occurrence is hidden (i.e. single select unselected option)
+        // so check the style directly
+        return this.formSetOccurrencesContainer?.getEl().getHTMLElement().style.display !== 'none';
```
alansemenov, line 244: `this.formSetOccurrencesContainer?.getEl().getHTMLElement().style.display -> this.formSetOccurrencesContainer?.getEl().getDisplay()`

src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
```diff
@@ -225,13 +229,19 @@ export abstract class FormSetOccurrenceView
     }
 
     showContainer(show: boolean) {
```
alansemenov, line 231: Rename this to `setContainerVisible`, since it can also hide, not just show. Also rename "`show`" argument to "`visible`".

alansemenov (changes requested):

alansemenov (approved):

Issue #1685, opened 2021-01-05
No reset function for single-select option-set


alansemenov:
It should be possible to reset selected option in a single-select option-set and collapse/expand the form by clicking the header.

* Replace "Collapse" option in the context-menu with "Reset"
* "Reset" should be enabled only when an option is selected
* When an option is selected, the dropdown should look and behave like the header of a multi-select option-set. This can be achieved by removing border of the dropdown, dropdown handle button and setting `padding: 4px 13px 0 13px` on `single-selection-header` and "margin: 0" to `option-value`. In this mode we should also apply click handlers to `single-selection-header` so that it expands/collapses the form on click, just like header of a multi-select option-set does (note that there's a tooltip and different cursor on hover as well). If there's no underlying form, nothing should happen on click and there should be no hover effect.

![image](https://user-images.githubusercontent.com/11612490/103663211-24122600-4f71-11eb-8461-bd98024810e6.png)

* When "Reset" is clicked in the context menu, remove the form (if it exists), switch the mode (so that the dropdown looks like a dropdown again) and reset selection in the dropdown:

![image](https://user-images.githubusercontent.com/11612490/103664625-c383e880-4f72-11eb-8164-828b15d73316.png)

PR #1718, opened 2021-01-14
No reset function for single-select option-set #1685

pmi:
**NB:** Had to revert changes made in 3327dac89ef87a38b62aedc9911def81c1182d4a because that was an option ID which was saved to property set so label just could not be used there.


alansemenov (changes requested):
Works (almost) like a charm!

Two things:
1. Drag control remains visible after option is reset:
image

2. Selected option's title won't dynamically change based on the form's contents. This might be the same as https://github.com/enonic/lib-admin-ui/issues/1709.
image

PR #1742, opened 2021-01-18
Title of a single-select option-set occurrence is not updated dynamic…

pmi:
…ally #1738

**NB** This change also added real-time title update !


alansemenov (changes requested):
Title of the occurrence is not correct after an option is selected - it displays the entire set's label, not the selected option's as it used to be:

image

alansemenov (approved):

Issue #1788, opened 2021-02-09
Incorrect label resolution of an option-set inside an item-set


alansemenov:
When the first non-empty property value for an item-set is a selected option of an option-set, the title is resolved to the option's **name** (_option_1_) instead of label (_My first option_). Note that for the option-set itself it's resolved correctly.

![image](https://user-images.githubusercontent.com/11612490/107391156-b5d4fc00-6af8-11eb-859d-a71e74d18b4a.png)

```xml
    <item-set name="myItemSet">
      <label>ItemSet</label>
      <items>
        <option-set name="checkOptionSet">
          <label>Multi-selection OptionSet</label>
          <expanded>true</expanded>
          <occurrences minimum="1" maximum="1"/>
          <options minimum="1" maximum="3">
            <option name="option_1">
              <label>My first option</label>
            </option>
            <option name="option_2">
              <label>My second option</label>
            </option>
          </options>
        </option-set>
      </items>
      <occurrences minimum="1" maximum="1"/>
    </item-set>
```

PR #1799, opened 2021-02-16
Incorrect label resolution of an option-set inside an item-set #1788

pmi:


src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
```diff
@@ -436,21 +439,67 @@ export abstract class FormSetOccurrenceView
...
+                            .sort((one: Property, two: Property) => {
+                                return nameLabelMap.get(one.getString()).index - nameLabelMap.get(two.getString()).index;
+                            })
+                            .map((selectedProp: Property) => nameLabelMap.get(selectedProp.getString()).label);
+
+                        labels.push(...selectedLabels);
+
+                        if (labels.length === 0) {
+                            // should not happen, but in case no labels were found go inside selected options forms
+                            selectionArray.some((selectedProp) => {
```
alansemenov, line 468: Can we please refactor this?

```Typescript
if (formItem instanceof FormOptionSet) {
  this.fetchLabelFromOptionSet(propertySet);
elseif (formItem instanceof FormItemSet) {
  this.fetchLabelFromItemSet(propertySet);
} else {
  ...
}
```

It actually seems like in the end you always do

```Typescript
this.recursiveFetchLabels(arr, labels, firstOnly);
return firstOnly && labels.length > 0;

```
so maybe it can/should be simplified even more.

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/form/set/FormSetOccurrenceView.ts
pmi: Yep, will do that ✔️ 

No, not always ❌ 

😆

alansemenov (approved):

PR #1802, opened 2021-02-17
Responsive Wizard header

ashklianko:


alansemenov:
@ashklianko These changes will break app-users, so need a PR for that as well

ashklianko:
> @ashklianko These changes will break app-users, so need a PR for that as well

For sure, but first wanted you to verify that this change is appropriate

alansemenov:
Seems like a smart solution 👍

alansemenov (approved):

PR #1814, opened 2021-02-23
Option Set - uploaded image inside deselected option are not cleared …

pmi:
…after saving the changes #1758


alansemenov:
The fix should be option-set specific

Issue #1821, opened 2021-02-25
Input validation check is triggered multiple times


ashklianko:
For example in `TextInputType.isValid()` is invoked at least 3 times when validity is not changed after value change and 5-8!!! times if validity changed

PR #1853, opened 2021-03-15
Input validation check is triggered multiple times #1821

ashklianko:
Reworked validation for BaseInputTypeNotManagingAdd inheritors to avoid code duplication and redundant validation calls.
1. On user input validating only updated input instead of all occurrences.
2. Triggering validity change only when validity changed
3. Lots of small code updates


alansemenov (changes requested):
Codacy issues and conflicts here

ashklianko:
Updated

alansemenov (changes requested):
Still some Codacy issues... Seems like some of the new code has to be refactored for simplicity

image

alansemenov (approved):

PR #1867, opened 2021-03-31
Move cleanup of option-set's unselected options to FormOptionSetOccurrenceView

ashklianko:
-handling removing of empty selection  ('_selected' array and empty non selected props) right in option set


alansemenov (approved):
Very nice refactoring, @ashklianko. That's where the cleanup should be  👍

PR #1935, opened 2021-05-03
Move content-related classes to Content Studio

alansemenov:


ashklianko (approved):
Great! At last things in places they supposed to be

PR #1944, opened 2021-05-11
Validation in the Insert Link dialog won't allow 1 or 2 char long domains

alansemenov:


sigdestad:
Any reason why we should not allow 1-character domains as well? Ref forum posting...

sigdestad:
Also, then there is internal links, where people might even use domains without TLD i.e. http://intranet/whatever

ashklianko (approved):

PR #1994, opened 2021-06-12
Increase height of selector dropdowns #1993

ashklianko:
-Height of a slickgrid's row must not be set via css/less but via internal slickgrid rowHeight param, we have setRowHeight in grid options object for it
-Using display: flex for grid cells and its' children elements wherever possible
-Moved common slick-cell styling to a grid.less and removed duplicated entries in other places
-Added separate names-and-icon-viewer.less for NamesAndIconViewer.ts styling
-Adjusted corresponding selected-options styling


alansemenov (changes requested):
1. Something happened with alignment
image

2. In the tree mode when you expand a node with children, the expanded node will always be highlighted as selected:

image
image

alansemenov (approved):

PR #2076, opened 2021-08-04
Migrate Admin API from XP runtime

pmi:
Depends on https://github.com/enonic/xp/pull/8994


rymsha:
Maybe move these resources to CS directly?

alansemenov (approved):

PR #2137, opened 2021-09-20
Support adding new entries to localisation resources

ashklianko:


src/main/resources/assets/admin/common/js/util/Messages.ts
```diff
@@ -24,6 +24,16 @@ export class Messages {
...
 
+    static addMessages(messages: Object) {
+        if (messages) {
+            for (let key in messages) {
+                if (messages.hasOwnProperty(key)) {
+                    Messages.getMessages().set(key, messages[key]);
+                }
+            }
+        }
+    }
```
alansemenov, line 35: This is almost exactly the same as `setMessages`... Can't we reuse that method?


```
    static setMessages(messages: Object, clear: boolean = true) {
        if (messages) {
            if (clear) Messages.getMessages().clear();
            for (let key in messages) {
                if (messages.hasOwnProperty(key)) {
                    Messages.getMessages().set(key, messages[key]);
                }
            }
        }
    }
```

Or the other way around, keep `addMessages` as it is now and change `setMessages` like this:

```
    static setMessages(messages: Object) {
        if (!messages) return;

        Messages.getMessages().clear();
        Messages.addMessages(messages);
    }
```

alansemenov (changes requested):

alansemenov (approved):

Issue #2163, opened 2021-09-28
Server event: add newPath param to the NodeServerChangeItem class


ashklianko:
Now this param is received in server event via websocket but not set to a NodeServerChangeItem class vars

PR #2164, opened 2021-09-28
Server event: add newPath param to the NodeServerChangeItem class #2163

ashklianko:


ashklianko:
Also required for Archive tasks

alansemenov (approved):

PR #2224, opened 2021-10-29
Support callback function in Notification actions

pmi:


src/main/resources/assets/admin/common/js/notify/NotificationMessage.ts
```diff
@@ -12,10 +12,9 @@ export class NotificationMessage
         super('notification');
         this.autoHide = autoHide;
         const notificationInner = new DivEl('notification-inner');
-        const notificationRemove = new SpanEl('notification-remove');
-        notificationRemove.setHtml('X');
+        const notificationRemove = new SpanEl('notification-remove icon-close');
         this.notificationContent = new DivEl('notification-content');
-        this.notificationContent.setHtml(message);
+        this.notificationContent.setHtml(message, false);
```
alansemenov, line 17: We cannot just quietly remove escaping, it was added for a reason. I suggest doing this differently: add two new properties for passing a link text and a callback function. If both these properties are set, append a link to the notification text which will execute a specified callback on click.

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/notify/NotificationMessage.ts
pmi: Refactored addAction() method to append both text to the notification body and the click handler

alansemenov (approved):

PR #2263, opened 2021-11-12
Support custom validation errors for input types #2262

ashklianko:
-Added isValidationErrorToBeRendered to InputTypeView to move InputView related logic (to render or not render validation error) to it. 
-Cleaned InputValidationRecording from validationErrorToBeRendered and validationMessageToBeRendered that are totally do not belong to this class
-Added hideValidationDetailsByDefault to InputTypeView to let input type decide rather to show or not validation details under each input by default
-Updated validation details toggler
-Showing custom validation errors for each input occurrence
-Using styles wherever possible to handle show/hide of validation

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/InputView.ts
+++ b/src/main/resources/assets/admin/common/js/form/InputView.ts
@@ -24,7 +24,7 @@ import {InputLabel} from './InputLabel';
 import {InputTypeManager} from './inputtype/InputTypeManager';
 import {ValidationRecordingPath} from './ValidationRecordingPath';
 import {InputViewValidationViewer} from './InputViewValidationViewer';
-import {ValidationError} from '../ValidationError';
+import {TogglerButton} from '../ui/button/TogglerButton';
 
 export interface InputViewConfig {
 
@@ -41,6 +41,8 @@ export class InputView
     extends FormItemView {
 
     public static debug: boolean = false;
+    private static ERROR_DETAILS_HIDDEN_CLS: string = 'error-details-hidden';
+
     private input: Input;
     private parentPropertySet: PropertySet;
     private propertyArray: PropertyArray;
@@ -48,6 +50,7 @@ export class InputView
     private bottomButtonRow?: DivEl;
     private addButton?: Button;
     private validationViewer: InputViewValidationViewer;
+    private validationDetailsToggler: TogglerButton;
     private previousValidityRecording: ValidationRecording;
     private validityChangedListeners: { (event: RecordingValidityChangedEvent): void }[] = [];
     private helpText?: HelpTextContainer;
@@ -65,7 +68,6 @@ export class InputView
 
         this.input = config.input;
         this.parentPropertySet = config.parentDataSet;
-
     }
 
     public layout(validate: boolean = true): Q.Promise<void> {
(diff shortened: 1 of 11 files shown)


src/main/resources/assets/admin/common/js/form/inputtype/support/BaseInputTypeNotManagingAdd.ts
```diff
@@ -357,6 +341,16 @@ export abstract class BaseInputTypeNotManagingAdd
             const inputEl: Element = occurrenceView.getInputElement();
             this.validateUserInput(inputEl);
             this.validateRequiredContract(inputEl, value);
+            this.validateCustomErrors(occurrenceView);
         }
     }
+
+    protected validateCustomErrors(occurrenceView: InputOccurrenceView): void {
+        this.getContext().formContext.getValidationErrors().forEach((error: ValidationError) => {
+            if (occurrenceView.getDataPath().asRelative().toString() === error.getPropertyPath()) {
```
alansemenov, line 350: `occurrenceView.getDataPath().asRelative().toString()` should be calculated once, outside the forEach loop.

src/main/resources/assets/admin/common/js/form/inputtype/support/BaseInputTypeNotManagingAdd.ts
```diff
@@ -357,6 +341,16 @@ export abstract class BaseInputTypeNotManagingAdd
...
+                this.occurrenceValidationState.get(occurrenceView.getInputElement().getId()).addAdditionalValidation(
```
alansemenov, line 351: Same with `this.occurrenceValidationState.get(occurrenceView.getInputElement().getId())`

src/main/resources/assets/admin/common/js/form/InputView.ts
```diff
@@ -117,15 +118,45 @@ export class InputView
                 this.bottomButtonRow = new DivEl('bottom-button-row');
                 this.appendChild(this.bottomButtonRow);
                 this.bottomButtonRow.appendChild(this.addButton);
+
+                this.toggleClass('one-and-only',
+                    this.input.getOccurrences().getMinimum() === 1 && this.input.getOccurrences().getMaximum() === 1);
+
+                inputTypeViewNotManagingAdd.onOccurrenceValueChanged(() => {
+                    this.toggleClass('has-invalid-user-input', !inputTypeViewNotManagingAdd.hasValidUserInput());
```
alansemenov, line 126: This should be made into a separate method since it's used in 3 different handlers.

src/main/resources/assets/admin/common/js/form/InputView.ts
```diff
@@ -117,15 +118,45 @@ export class InputView
                 this.bottomButtonRow = new DivEl('bottom-button-row');
                 this.appendChild(this.bottomButtonRow);
                 this.bottomButtonRow.appendChild(this.addButton);
+
+                this.toggleClass('one-and-only',
```
alansemenov, line 122: Rename the css class to something less romantic, like "`single-occurrence`".

alansemenov (changes requested):
A **BIG** problem: validation on save is now one step behind. If you fill in a blank input with an invalid value and save, content will remain valid and no validation error will be shown until the page is refreshed. Same if you change an invalid value with a valid one and save - the content will remain invalid and error message won't disappear until you refresh the page.

alansemenov (approved):

Issue #2275, opened 2021-11-19
Input with a mix of valid and invalid occurrences is considered valid


ashklianko:
@alansemenov commented on [Mon Nov 15 2021](https://github.com/enonic/app-contentstudio/issues/3834)

When there's a mix of valid and invalid occurrences in the same input view, the input is considered valid:

![image](https://user-images.githubusercontent.com/11612490/141769812-416e0325-04a4-443a-a9c8-59017d0089c1.png)

PR #2276, opened 2021-11-19
Input with a mix of valid and invalid occurrences is considered valid…

ashklianko:
… #2275

-Setting error message for custom errors in occurrences so input becomes invalid when custom error presents

First I thought of not putting any message for Input validation viewer since input occurrences with failed server validation  has a personal error message under it, however it was not clear why input is invalid since it has enough valid occurrences, so I've added special message indicating why input is still invalid with enough valid occurrenes

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/AdditionalValidationRecord.ts
+++ b/src/main/resources/assets/admin/common/js/form/AdditionalValidationRecord.ts
@@ -1,9 +1,12 @@
 export class AdditionalValidationRecord {
 
-    private message: string;
+    private readonly message: string;
+
+    private readonly custom: boolean;
 
     constructor(builder: Builder) {
         this.message = builder.message;
+        this.custom = builder.custom || false;
     }
 
     public static create(): Builder {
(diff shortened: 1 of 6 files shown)


alansemenov (changes requested):
* There's still an issue when a mix of invalid and valid occurrences makes the input and the content is valid. It happens when _another_ invalid input becomes valid.

Here "Required Long" is invalid (and "Long" is marked as invalid too):
image

Here "Required Long" becomes valid and for some reason this makes "Long" (and the entire content) valid as well:
image

* There should be no "_Hide details_" link without error message. If there's one occurrence then we should show errors right away:

image

* We should get rid of "_Custom validation errors_". I don't understand the logic behind showing it and it bears no meaning for the user. It doesn't matter if an input is invalid because of a custom or a built-in validation. What matters is if an occurrence has an error, in which case we also show the general validation error. In case of custom validation, we should show exactly the same general validation errors that we show for built-in validation.

Here are the use-cases:
1) One occurrence (min=1, max=1), validation error (doesn't matter custom or not)

Now:
image

Should be (no general validation error, occurrence error is shown right away):
image

2) Multiple occurrences (min>=1, max>1), validation error

Now:
image

Should be (occurrence errors are hidden by default, general validation is shown with "Show Details" link):
image

For some reason the last use case is correct when both occurrences has custom errors, so I don't understand the logic here at all. Anyway, it should be the same error message as for the built-in validation.

ashklianko:
Updated

Issue #2284, opened 2021-11-25
Toolbar: let actions manually fold/unfold


ashklianko:
In certain cases need to force all toolbar actions to be fold in one dropdown.
Also letting manually prevent resposive fold/unfold

PR #2285, opened 2021-11-25
Toolbar: let actions manually fold/unfold #2284

ashklianko:
-also small fix for panel

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/toolbar/Toolbar.ts
+++ b/src/main/resources/assets/admin/common/js/ui/toolbar/Toolbar.ts
@@ -187,7 +187,7 @@ export class Toolbar
         this.foldButton.setLabel(this.areAllActionsFolded() ? i18n('action.actions') : i18n('action.more'));
     }
 
-    setFoldButtonLabel(value: string) {
+    setFoldButtonLabel(value: string): void {
         this.foldButton.setLabel(value);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/ui/toolbar/Toolbar.ts
```diff
@@ -14,6 +14,7 @@ export class Toolbar
 
     protected fold: FoldButton;
     protected actions: Action[] = [];
+    protected locked: boolean;
```
alansemenov, line 17: Why protected? If you need to access it from the outside, implement `isLocked` method.

src/main/resources/assets/admin/common/js/ui/toolbar/Toolbar.ts
```diff
@@ -83,41 +89,55 @@ export class Toolbar
...
+            this.doFold();
+        } else {
+           this.doExpand();
+        }
+
+        this.updateFoldButtonLabel();
+    }
 
-            while (toolbarWidth <= this.getVisibleButtonsWidth() && this.getNextFoldableButton()) {
+    doFold(force: boolean = false): void {
```
alansemenov, line 105: Rename to `fold`

src/main/resources/assets/admin/common/js/ui/toolbar/Toolbar.ts
```diff
@@ -83,41 +89,55 @@ export class Toolbar
...
-                let buttonToShow = this.fold.pop();
-                buttonToShow.insertBeforeEl(this.fold);
+    protected getToolbarWidth(): number {
+        return this.getEl().getWidthWithoutPadding();
+    }
 
-                if (this.fold.isEmpty()) {
-                    this.fold.hide();
-                }
+    doExpand(): void {
```
alansemenov, line 127: Rename to `expand`

alansemenov (changes requested):

alansemenov (approved):

Issue #2317, opened 2021-12-17
Allow space distribution in SplitPanel without firing resize event


ashklianko:
If we use a splitter between panels in SplitPanel to change its sizes then no need to notify entire page that available size changed because it did not, use ResponsiveItem api to update those items that have sizes changed

PR #2318, opened 2021-12-17
SplitPanel: allow space distribute without firing ResponsiveManager.f…

ashklianko:
…ireResizeEvent() #2317

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/panel/SplitPanel.ts
+++ b/src/main/resources/assets/admin/common/js/ui/panel/SplitPanel.ts
@@ -334,41 +334,23 @@ export class SplitPanel
 
     distribute(noResizeEvent: boolean = false): void {
         if (this.isHorizontal()) {
-            this.firstPanel.getEl().setHeight(this.getPanelSizeString(1)).setWidth(null);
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/admin/common/js/ui/panel/SplitPanel.ts
```diff
@@ -327,20 +332,30 @@ export class SplitPanel
...
             this.splitter.getEl().setHeightPx(this.getSplitterThickness()).setWidth(null).setLeft(null);
             if (this.isVisible()) {
-                this.triggerResizeDebounced();
+                if (noResizeEvent) {
+                    this.firstPanelResponsiveItem.update();
+                    this.secondPanelResponsiveItem.update();
+                } else {
+                    this.triggerResizeDebounced();
+                }
             }
```
alansemenov, line 347: This piece of code is duplicated 10 lines below. Extract to `doAfterDistribute` or something?

alansemenov: Or maybe move this block after `if.. then..`, unless it has to be called before setting the splitter position.

alansemenov (approved):

Issue #2319, opened 2021-12-17
(Accessibility) modal dialogs don't prevent keyboard navigation in background elements


reisfmb:
Related issues in content studio:
- [#2984](https://github.com/enonic/app-contentstudio/issues/2984)
- [#2990](https://github.com/enonic/app-contentstudio/issues/2990)

reisfmb:
Fixed in https://github.com/enonic/lib-admin-ui/issues/2321

PR #2320, opened 2021-12-17
Added focus trap to ModalDialog (#2319)

reisfmb:
This is a quick accessibility fix using the lib [focus-trap](https://github.com/focus-trap/focus-trap). 

If accepted it'll trap the tab navigation to tabbable elements within the modal dialog, and therefore for all modals that extends that class (`NewContentDialog` in the app-contentstudio for example).

This would fix some content studio accessibility issues listed [here](https://github.com/enonic/app-contentstudio/issues/2988).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "dompurify": "^2.3.4",
     "fine-uploader": "^5.16.2",
     "flag-icon-css": "^3.5.0",
+    "focus-trap": "^6.7.1",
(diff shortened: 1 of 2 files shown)


ashklianko:
Hi Bruno! We already handle focus in/out in ModalDialog, however as you may found there's a flaw in it preventing to switch to the first tabbable element in some cases. I've created a task #2321 to fix it

PR #2332, opened 2021-12-21
Implement configurable position of dropdown handler in MenuButton

ashklianko:
-AppHelper: set a last focused element in focusout callback
-ModalDialog:
1) Update list of tabbable items on show and when new elements added into a dialog
2) On focusout check correctly focus last or first tabbable
-MenuButton: define dropdown handle position (right or left) to correctly navigate through it

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
+++ b/src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
@@ -5,40 +5,53 @@ import {Action} from '../Action';
 import {DropdownHandle} from './DropdownHandle';
 import {ActionButton} from './ActionButton';
 import {Body} from '../../dom/Body';
+import {Element} from '../../dom/Element';
+import * as Q from 'q';
+
+export enum MenuButtonDropdownPos {
+    LEFT, RIGHT
+}
 
 export class MenuButton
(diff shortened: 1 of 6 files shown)


ashklianko:
That task will help with keyboard navigation epic #2988

src/main/resources/assets/admin/common/js/ui/button/MenuButton.ts
```diff
@@ -125,19 +138,23 @@ export class MenuButton
...
 
-    setToggleMenuOnAction(value: boolean) {
+    setToggleMenuOnAction(value: boolean): void {
         this.toggleMenuOnAction = value;
     }
 
+    protected getDropdownPosition(): MenuButtonDropdownPos {
+        return MenuButtonDropdownPos.LEFT;
+    }
+
```
alansemenov, line 157: This is a bit strange. Before the changes it used to be like this:

`let children = [this.actionButton, this.dropdownHandle, this.menu];`

with `dropdownHandle` being **to the right** of the `actionButton`. However, with this new method default position of the `dropdownHandle` is `to the left`. It's important that we don't break anything with these changes, in case MenuButton was used elsewhere.

alansemenov (approved):

Issue #408, opened 2018-03-14
Use IntersectionObserver for lazy rendering


alansemenov:
Our lazy-rendering of complex field-sets is now using ``getBoundingClientRect()`` to track when an element is scrolled into the viewport (see ``Element.lazyRender`` and ``Element.isInViewPort``). Replace it with much faster ``IntersectionObserver`` which has now landed in all major browsers.

https://caniuse.com/#search=intersection
https://developers.google.com/web/updates/2016/04/intersectionobserver

PR #2342, opened 2021-12-28
Implement LazyListBox based on IntersectionObserver

ashklianko:


alansemenov (approved):
Finally! I've created a [task](https://github.com/enonic/lib-admin-ui/issues/408) for migrating to `IntersectionObserver` for lazy loading almost 4 years ago and now it's finally at a point where we can use it overall. We should create tasks for using it in TreeGrid as well.

PR #2343, opened 2021-12-29
reset button on date pickers

poi33:
Added a reset button to the date & datetime picker
Resets to the default value of the input value or no input (empty string)


alansemenov:
@poi33 have updated the task description

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -12,16 +12,24 @@ import {SelectedDateChangedEvent} from './SelectedDateChangedEvent';
 export class Picker<T extends Element>
     extends DivEl {
 
+    private showDefaultButton: Boolean = false;
```
alansemenov, line 15: `boolean` is a primitive type. But you don't need this extra property, `defaultValueHandler` is enough.

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -47,6 +55,13 @@ export class Picker<T extends Element>
         this.input.resetBaseValues();
     }
 
+    public setDefaultHandler(setDefault: Function, defaultDate: Date | undefined): void {
```
alansemenov, line 58: Rename method to `setDefaultValueHandler`, rename `setDefault` argument to `handler`, remove `defaultDate`

src/main/resources/assets/admin/common/js/ui/time/DatePicker.ts
```diff
@@ -45,6 +52,15 @@ export class DatePicker
...
             this.setDate(builder.date);
         }
+
+        if (builder.defaultDate) {
+            this.setDefaultHandler(
+                () => {
+                    this.setSelectedDate(builder.defaultDate);
+                },
+                builder.defaultDate
+            );
```
alansemenov, line 62: You don't need to pass both handler AND the value the handler sets. The handler is enough. And curly brackets in the handler are not needed.

if (builder.defaultDate) {
    this.setDefaultHandler(() => this.setSelectedDate(builder.defaultDate));
}

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -12,16 +12,24 @@ import {SelectedDateChangedEvent} from './SelectedDateChangedEvent';
 export class Picker<T extends Element>
     extends DivEl {
 
+    private showDefaultButton: Boolean = false;
+
     protected popup: T;
 
     protected popupOkButton: Button;
 
+    protected popupDefaultButton: Button;
```
alansemenov, line 21: We don't need this as a class property, just like we don't need `popupOkButton` as a class property. `init` methods return an instance of each button which is then used inside `createPopup`.

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -12,16 +12,24 @@ import {SelectedDateChangedEvent} from './SelectedDateChangedEvent';
...
+
     protected selectedDate: Date;
 
     protected input: TextInput;
 
+    protected defaultDate: Date;
```
alansemenov, line 27: Not needed.

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -47,6 +55,13 @@ export class Picker<T extends Element>
         this.input.resetBaseValues();
     }
 
+    public setDefaultHandler(setDefault: Function, defaultDate: Date | undefined): void {
+        if (defaultDate) {
+            this.showDefaultButton = true;
+        }
```
alansemenov, line 61: Remove

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -12,16 +12,24 @@ import {SelectedDateChangedEvent} from './SelectedDateChangedEvent';
...
+
     protected selectedDate: Date;
 
     protected input: TextInput;
 
+    protected defaultDate: Date;
+
     protected validUserInput: boolean;
 
+    private defaultHandler: Function;
```
alansemenov, line 31: Rename to `defaultValueHandler`

src/main/resources/assets/admin/common/js/ui/time/DateTimePicker.ts
```diff
@@ -72,6 +79,15 @@ export class DateTimePicker
...
         }
+
+        if (builder.defaultDate) {
+            this.setDefaultHandler(
+                () => {
+                    this.setSelectedDateTime(builder.defaultDate);
+                },
+                builder.defaultDate
+            );
+        }
```
alansemenov, line 90: See comment for DatePicker.

A side note: we should refactor DateTimePicker to extend/inherit from DatePicker to eliminate code duplication. Same with their builders.

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -187,13 +202,25 @@ export class Picker<T extends Element>
         }
     }
 
+    private initDefaultButton() {
+        this.popupDefaultButton = new Button(i18n('action.setDefault'));
+        this.popupDefaultButton.addClass('default-button');
+        this.popupDefaultButton.onClicked(() => {
+            if (this.setDefaultHandler) {
+                this.defaultHandler();
+            }
```
alansemenov, line 211: Condition `if (this.setDefaultHandler) `doesn't make sense. This method is defined in the class, so it always exists.

````
private initDefaultValueButton() {
  ...
  const defaultValueButton = ...;
  defaultValueButton.onClicked(() => this.defaultValueHandler());
  return defaultValueButton;
}

File: src/main/resources/assets/admin/common/js/ui/time/Picker.ts [hunk 10]
```diff
@@ -203,7 +230,16 @@ export class Picker<T extends Element>
 
         this.initPopup(this.builder);
         this.setupPopupListeners(this.builder);
-        this.initCloseButton();
+        const buttonContainer = new DivEl('btn-container');
+
+        if (this.showDefaultButton) {
+            buttonContainer.appendChild(this.initDefaultButton());
+        }
```
alansemenov, line 237:
```
if (this.defaultValueHandler) {
    buttonContainer.appendChild(this.initDefaultButton());
}
```

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -203,7 +230,16 @@ export class Picker<T extends Element>
...
+
+        const okButton = this.initCloseButton();
```
alansemenov, line 239: `closeButton` should not be called `okButton`.

alansemenov (changes requested):
Also, the new button should be grey and left-aligned in the dropdown:

image

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
ashklianko: ...and type definition is also required

poi33: Refactored the ok button to close button. 😃

alansemenov (changes requested):
Almost there, just tiny things left. Can you also please disable the "Set default" button if current value = default? There should be a case when user clicks a button and nothing happens.

1. Make the button a class property in Picker.ts (but call it properly, `defaultValueButton`).
2. Originally disable the button (`this.defaultValueButton.setEnabled(false)`).
3. Implement a new listener for `onSelectedDateTimeChanged` and enable the button whenever the date/time changes. 
4. Disable it again after calling `this.defaultValueHandler` in `onClicked`.

poi33:
@ashklianko @alansemenov updated 
Needed additional changes, as the onSelectedDateTimeChanged did not trigger on the Calender date change.

src/main/resources/assets/admin/common/js/ui/time/Picker.ts
```diff
@@ -8,24 +8,41 @@ import {AppHelper} from '../../util/AppHelper';
...
+import {DateTime} from '../../util/DateTime';
+
+/**
+ * Not required for pickers
+ * Should only be used when th default value is needed in children
+ */
+export class DefaultValuePickerBuilder {
+    defaultValue: any;
+
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
poi33, line 20: This seems like a place that we want to have `any`  since we could use the picker for any value in the future

alansemenov:
Should be closed and redone after refactoring of DatePicker classes

Issue #2346, opened 2022-01-03
Support required checkbox


alansemenov:
Since Checkbox input type doesn't support occurrences, it's currently not possible to implement a checkbox that must be checked off.

Implement a special case of validation for Checkbox which will be triggered when `occurrences.minimum > 0 ` and will display "_This field is required_" (same validation message as for occurrence-based input types) message if the checkbox is not checked off.

Initially described in https://github.com/enonic/lib-admin-ui/issues/2344

PR #2349, opened 2022-01-03
Support required checkbox #2346

ashklianko:
-if min occurrences is more than 0 then validating checkbox: is valid if checkbox is checked and not valid if opposite


alansemenov (approved):

alansemenov:
Works as intended 👍

PR #2354, opened 2022-01-04
LazyListBox: let scrollable container property be overriden by inheri…

ashklianko:
…tors #2353

-refactoring


alansemenov (approved):
Yes!

PR #2355, opened 2022-01-10
New ItemSet occurrence gets instantly validated

ashklianko:


ashklianko:
-add occurrence buttons are using 'createAndAddOccurrence' method, I've updated it to use css to hide validation error immediately after occurrence is created and validated, and show it on next call to validate()
-Updated app config dialog to override form styling and force showing  validation error for newly added occurrence

alansemenov (approved):

PR #2369, opened 2022-01-17
Implement ConfigLoader

alansemenov:


src/main/resources/assets/admin/common/js/util/Config.ts
```diff
@@ -0,0 +1,99 @@
...
+    private static getPropertyValue(property: string): JSONValue {
+        if (property.indexOf('.') > 0) {
+            return CONFIG.getNested(property);
+        }
+        if (!CONFIG.has(property)) throw `Config property ${property} not found`;
+        return CONFIG.CACHE[property];
+    }
+
+    static setConfig(config: JSONObject): void {
+        CONFIG.CACHE = Object.freeze(Object.assign({}, config));
```
ashklianko, line 51: Is there a need to let anyone override config's cache?

src/main/resources/assets/admin/common/js/util/Config.ts
```diff
@@ -0,0 +1,99 @@
...
+    static init(url: string): Q.Promise<JSONObject> {
+        return CONFIG.loadAndCache(url);
+    }
+
+    static isTrue(property: string): boolean {
+        return CONFIG.get(property) === 'true';
+    }
+
+    static getConfig(): JSONObject {
+        return CONFIG.CACHE;
```
ashklianko, line 19: Do we need to expose CONFIG.CACHE to outer world?

ashklianko (approved):

PR #2390, opened 2022-01-25
Create head request #2389

pmi:


src/main/resources/assets/admin/common/js/rest/HeadRequest.ts
```diff
@@ -0,0 +1,29 @@
...
+    extends Request {
+
+    constructor() {
+        super(HttpMethod.HEAD);
+    }
+
+    protected prepareRequest(): void {
+        super.prepareRequest();
+
+        this.request.setRequestHeader('Accept', 'application/json');
```
rymsha, line 16: Why would HEAD request only accept `application/json` ?

pmi: Well, because it is a copy from GetRequest which had that :) 
Agree that it is not necessary since head doesn't load body at all. 
It doesn't do any harm either since it is a hint to server what client is expecting, but not a requirement.

rymsha (approved):

alansemenov (approved):

PR #2402, opened 2022-02-03
Create method to show/hide tabs in DockedPanel #2401

pmi:


src/main/resources/assets/admin/common/js/ui/panel/DockedPanel.ts
```diff
@@ -41,6 +41,24 @@ export class DockedPanel
...
+                // currently visible item has been hidden, so find a substitute to show
+                for (let i = 0; i < this.items.length; i++) {
+                    const item = this.items[i];
+                    if (i !== index && !item.hasClass('hidden')) {
+                        this.navigator.selectNavigationItem(i);
+                        break;
+                    }
+                }
+            }
+        }
```
alansemenov, line 59: The logic here is very difficult to follow. Also, there's no point in using yet another CSS class (`"hidden"`) if we already have `hide`/`show` on every element.

So it can be simplified like this (has to be tested):

    setItemVisible<T extends Panel>(panel: T, visible: boolean): void {
        const currentIndex = this.deck.getPanelIndex(panel);
        if (currentIndex < 0) { // The panel is not found
            return;
        }

        const isCurrentlyVisible = this.deck.getPanelShownIndex() === currentIndex;
        if (isCurrentlyVisible === visible) {
            return;
        }

        this.items[currentIndex].setVisible(visible);
        if (isCurrentlyVisible) {
            const newItemToShow = this.items.find((item, index) => item.isVisible() && currentIndex !== index);
            if (newItemToShow) {
                newItemToShow.show();
            }
        }
    }

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/ui/panel/DockedPanel.ts
pmi: The reason of using 'hidden' instead of show/hide is that we need to be able to do a check if item is hidden:
- If we use regular show/hide, then isVisible() will return false for every tab if the whole panel is not currently visible.
- If we use 'hidden' class, then hasClass('hidden') will return real item state indifferently of the panel visibility.

But the logic can be simplified I guess ;)

alansemenov:
> If we use regular show/hide, then isVisible() will return false for every tab if the whole panel is not currently visible.

No, it won't. `isVisible()` will return `false` for a panel, but not for an item (tab) since it's visible even if its panel is not. Did you test it?

pmi:
> > If we use regular show/hide, then isVisible() will return false for every tab if the whole panel is not currently visible.
> 
> No, it won't. `isVisible()` will return `false` for a panel, but not for an item (tab) since it's visible even if its panel is not. Did you test it?

That is right when whole dockedPanel is visible, but it will not work if you put it in a dialog window for instance that is hidden

alansemenov (approved):

Issue #2403, opened 2022-02-03
Expose slickGrid in Grid class


reisfmb:
There's the need to add accessibility to the "Create New" dialog in app users as requested in this issue: [ #649](https://github.com/enonic/app-users/issues/649)

This issue is necessary in order to complete that app users accessibility issue.

Issue #649, opened 2018-06-27
Image src is incorrectly stored in HTML Area content


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/41967157-44f91d58-7a01-11e8-9a1f-126913f73e94.png)

![image](https://user-images.githubusercontent.com/11612490/41967149-3cae88f4-7a01-11e8-9c42-375183744bf5.png)

PR #2404, opened 2022-02-03
Added getSlickGrid method in Grid class (#2403)

reisfmb:
This is related to the issue [#649](https://github.com/enonic/app-users/issues/649) in app users.


alansemenov:
We don't want to expose that

PR #2420, opened 2022-02-16
Part config with mandatory RadioButton within Option-Set (non-mandato…

ashklianko:
…ry) seem to struggle #2419

-Fix for npe in radiobutton related code

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/inputtype/radiobutton/RadioButton.ts
+++ b/src/main/resources/assets/admin/common/js/form/inputtype/radiobutton/RadioButton.ts
@@ -35,12 +35,11 @@ export class RadioButton
 
     layoutProperty(input: Input, property: Property): Q.Promise<void> {
         this.input = input;
-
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/admin/common/js/form/inputtype/radiobutton/RadioButton.ts
```diff
@@ -99,25 +98,25 @@ export class RadioButton
     }
 
     private readConfig(inputConfig: { [element: string]: { [name: string]: string }[]; }): void {
-        let options: { label: string; value: string; }[] = [];
+        const options: { label: string; value: string; }[] = [];
```
alansemenov, line 101: We should declare this as a type as it's used throughout the class.

ashklianko: Updated

alansemenov (approved):

PR #2422, opened 2022-02-17
Correctly resolve adminUri from config

alansemenov:


ashklianko (approved):

alansemenov:
> Is there a pull for CS? There's a code in `main.ts` using `UriHelper.getAdminUri('tool')`

Good catch! Here's PR for CS: https://github.com/enonic/app-contentstudio/pull/4256

PR #2467, opened 2022-03-14
Fix lazy loading in BaseRichCombobox

alansemenov:
* Calculate scrolled distance inside dropdown on `loaded` and `scroll` events instead of using `setInterval`
* Don't trigger `loaded` event in `BaseLoader` if data hasn't changed
* Support forced reload in `BaseRichComboBox`


ashklianko:
Made some updates

PR #2474, opened 2022-03-18
Custom Selector gives an error in post load mode

alansemenov:


ashklianko:
Seems to work fine, however when opening Edit Principals dialog with BaseRichCombobox it seems to fetch same data twice on FIRST dropdown expand:
image

ashklianko:
In general logic with postload seems to be overcomplicated, I'd say we need to refactor it more

ashklianko (approved):

ashklianko:
Updated

Issue #2510, opened 2022-04-12
Inner sortable triggers outer one


reisfmb:
This is a fix for cs issue: [#3982](https://github.com/enonic/app-contentstudio/issues/3982)

PR #2511, opened 2022-04-12
Added cancel prop to sortable of FormSetView (#2510)

reisfmb:
https://api.jqueryui.com/sortable/#option-cancel


reisfmb:
https://api.jqueryui.com/sortable/#option-cancel

ashklianko (approved):

PR #2531, opened 2022-04-26
Added keyboard accessibility to AppIcon

reisfmb:
@ashklianko 

Does it makes sense to let the app icon be reachable by keyboard if it does contain any action? 

This PR allows that. It was the behavior requested by Thomas Heartman in this [issue](https://github.com/enonic/app-users/issues/647)

If we don't want to allow that, i.e, only make app icon reachable by keyboard if there is an action attached to it, then we'll need to move line 20 to inside the conditional on line 22.

### DIFF
--- a/src/main/resources/assets/admin/common/js/app/bar/AppBar.ts
+++ b/src/main/resources/assets/admin/common/js/app/bar/AppBar.ts
@@ -45,7 +45,11 @@ export class AppBar
         return [this.showAppLauncherAction];
     }
 
-    setHomeIconAction() {
+    setHomeIconAction(): void {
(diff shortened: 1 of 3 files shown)


alansemenov:
> Does it makes sense to let the app icon be reachable by keyboard if it does contain any action?

@reisfmb No, we shouldn't make app icon tabbable if it doesn't contain any action. Actually, I'd like it to not have any hover effect if it doesn't do anything (like when no tabs are open in the Users app). Only when a tab (or tabs) are open, then the icon should become clickable and tabbable and get hover effect.

alansemenov (approved):

PR #2581, opened 2022-05-23
Limit the list of filter options in each aggregation #2580

ashklianko:
-added container for buckets with toggler button to show more/less buckets
-filtering out buckets with zero doc count


alansemenov (approved):
Good stuff!

Issue #2594, opened 2022-05-24
Help text and counter overlap for TextLine input


alansemenov:
When a TextLine input has both `help-text` and `config.showCounter`, then help text and counter will overlap:
![image](https://user-images.githubusercontent.com/11612490/170132679-05ab09ff-64fc-4245-8e36-d71f594d3db0.png)

Solution is to apply right padding to the help-text div when it's preceded by an input with both `text-input-type` and `show-counter` styles applied. Also check if we have the same issue with TextArea.

<img width="800" alt="image" src="https://user-images.githubusercontent.com/11612490/170136203-41b3d642-73bb-4377-a687-fdd623564fee.png">

jsi:
This bug is fixed for `show-counter`, but not for `max-length`.  We need a fix for max length as well.  See discussion with Tom Arild on Slack today (comments on my announcement of the release).

reisfmb:
@jsi Thank you. Will fix it today

https://github.com/enonic/lib-admin-ui/issues/2598

PR #2595, opened 2022-05-25
Help Text right padding if show counter is true (#2594)

reisfmb:
@alansemenov 

is it ok to use setTimeout as I did in this case? Doesn't seems right, but does the trick. 

The thing is that I need to execute `adjustHelpTextPadding` only after not only the parents are rendered, but also the child ones. What's the best way to do that?

Also, 100px of right padding is enough space for a counter of `<= 9999` characters, because 5 digits would overlap again. Should we take care of 5 digits?

### DIFF
--- a/src/main/resources/assets/admin/common/styles/api/input-common.less
+++ b/src/main/resources/assets/admin/common/styles/api/input-common.less
@@ -177,6 +177,12 @@
         }
       }
 
+      &.show-counter {
+        + .help-text {
+          padding-right: 100px;
+        }
+      }
+
       &.max-length-limited.show-counter {
         .separator {
           display: inline;


alansemenov (changes requested):
No no. As I said before, we should never apply any inline styling in the code, except for very rare cases. Here it's a simple CSS fix which I described in the issue description. If you look at the DOM, you'll see that div with `text-input-type` class (which is wrapper div for the input) is directly followed by the div with `help-text` class. So, when a div with `text-input-type` _also_ has `show-counter` class, then the adjacent div with `help-text` class should have padding-right applied to it.

Something like

```CSS
.text-input-type {
  &.show-counter + .help-text {
    padding-right: 50px;
  }
}
```
https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator

image

alansemenov (approved):

Issue #2598, opened 2022-05-30
Help text and max length counter overlap


reisfmb:
Need to fix for textline and textarea:

![image](https://user-images.githubusercontent.com/67838246/171031591-57133eed-a743-46d5-b659-7c6fcedce45b.png)

reisfmb:
@alansemenov 

Quick fix would be:
```css
&.max-length-limited {
    + .help-text {
        padding-right: 175px;
    }
}
```
I've used `175px` because it seems to be enough right padding for English text. The problem with this approach is that `175px` might not be the best depending on the language... For example, is too much for `160 tegn igjen`:

![image](https://user-images.githubusercontent.com/67838246/171057014-cec94afb-1590-438d-834f-9a4e70647cb1.png)

How should we proceed?

alansemenov:
> The problem with this approach is that 175px might not be the best depending on the language... For example, it is too much for 160 tegn igjen:

It's not a problem that it's too much. It's too difficult to guess or calculate width of the counter element and not worth spending time on.

PR #2605, opened 2022-05-31
Help text tight padding for max-length (#2598)

reisfmb:


src/main/resources/assets/admin/common/styles/api/input-common.less
```diff
@@ -183,6 +183,12 @@
...
 
+      &.max-length-limited {
+        + .help-text {
+          padding-right: 175px;
+        }
+      }
+
       &.max-length-limited.show-counter {
         .separator {
           display: inline;
```
alansemenov, line 194: As you can see, there's already selector for `&.max-length-limited` below, so you need to group the two:

```CSS
&.max-length-limited {
  + .help-text {
    padding-right: 175px;
  }

  &.show-counter {
    .separator {
       display: inline;
    }
  }
}
```

alansemenov (changes requested):

alansemenov (approved):

Issue #2612, opened 2022-06-06
Add keyboard navigation to SelectableListBoxDropdown


ashklianko:
@sgauruseu commented on [Wed Jun 01 2022](https://github.com/enonic/app-contentstudio/issues/4622)

1. Type an user name in "Last modified by" options filter input

**Issue**:  keyboard navigation doesn't work in the selector:

![image](https://user-images.githubusercontent.com/3728712/171348583-05bf6f18-b02c-4ab4-8336-58b2c9d2e37b.png)

PR #2619, opened 2022-06-06
SelectableListBoxDropdown: add keyboard navigation #2612

ashklianko:
-aligned behaviour with combobox


alansemenov (approved):
Great job!

PR #2642, opened 2022-06-21
Added validation for http, https, ftp and relative (#2641)

reisfmb:


src/main/resources/assets/admin/common/js/ui/form/Validators.ts
```diff
@@ -5,20 +5,31 @@ import {StringHelper} from '../../util/StringHelper';
...
-        return !regexUrl.test(value) ? i18n('field.value.invalid') : undefined;
+    public static validHttps(input: FormInputEl): string {
+        const regex = /^https:\/\/\w+(\.\w+)*(:[0-9]+)?((\/|\#).*)?(\?([^&=]+)=([^&=]+))?(?:&([^&=]+)=([^&=]+))*$/;
+        return !regex.test(input.getValue()) ? i18n('field.value.invalid') : undefined;
+    }
+
+    public static validHttp(input: FormInputEl): string {
+        const regex = /^http:\/\/\w+(\.\w+)*(:[0-9]+)?((\/|\#).*)?(\?([^&=]+)=([^&=]+))?(?:&([^&=]+)=([^&=]+))*$/;
+        return !regex.test(input.getValue()) ? i18n('field.value.invalid') : undefined;
+    }
```
alansemenov, line 24: Actually, I'm a bit sceptical to having two different methods for validating http and https. The regexps are identical (the only difference is the protocol) and it's too easy to make a mistake by changing the regexp in just one of the methods.

I think we should keep one method, as before (with a regexp that starts with `http(s)` prefix).

src/main/resources/assets/admin/common/js/ui/form/Validators.ts
```diff
@@ -5,20 +5,31 @@ import {StringHelper} from '../../util/StringHelper';
...
+
+    public static validFtp(input: FormInputEl): string {
```
alansemenov, line 26: Should be called `validFtpUrl`

src/main/resources/assets/admin/common/js/ui/form/Validators.ts
```diff
@@ -5,20 +5,31 @@ import {StringHelper} from '../../util/StringHelper';
...
+        const regex = /^ftp:\/\/[a-z0-9-]+(\.[a-z0-9-]+)+([/?].*)?$/;
+        return !regex.test(input.getValue()) ? i18n('field.value.invalid') : undefined;
+    }
+
+    public static validRelative(input: FormInputEl): string {
```
alansemenov, line 31: Should be called `validRelativeUrl`

src/main/resources/assets/admin/common/js/ui/form/Validators.ts
```diff
@@ -5,20 +5,31 @@ import {StringHelper} from '../../util/StringHelper';
...
 
     public static validEmail(input: FormInputEl): string {
-        let regexEmail = /\w+([-+.']\w+)*@\w+([-.]\w+)*\.\w+([-.]\w+)*/;
-        let value = input.getValue();
-        return !regexEmail.test(value) ? i18n('field.value.invalid') : undefined;
+        const regex = /\w+([-+.']\w+)*@\w+([-.]\w+)*\.\w+([-.]\w+)*/;
+        return !regex.test(input.getValue()) ? i18n('field.value.invalid') : undefined;
     }
 
-    public static validUrl(input: FormInputEl): string {
```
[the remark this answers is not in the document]
reisfmb: I only removed it because I couldn't find any other place in content studio using it... but yes, there might be other apps using it. My bad!

alansemenov (approved):

PR #2644, opened 2022-06-22
Can't reset search string in BaseLoader

alansemenov:
Compare search results with previously filtered dataset instead of the full one

### DIFF
--- a/src/main/resources/assets/admin/common/js/util/loader/BaseLoader.ts
+++ b/src/main/resources/assets/admin/common/js/util/loader/BaseLoader.ts
@@ -5,6 +5,7 @@ import {LoaderErrorEvent} from './event/LoaderErrorEvent';
 import {HttpRequest} from '../../rest/HttpRequest';
 import {ClassHelper} from '../../ClassHelper';
 import {Comparator} from '../../Comparator';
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/util/loader/BaseLoader.ts
```diff
@@ -92,15 +94,15 @@ export class BaseLoader<OBJECT> {
...
     search(searchString: string): Q.Promise<OBJECT[]> {
-        this.searchString = searchString;
+        this.setSearchString(searchString);
 
-        const searchResult: OBJECT[] = this.results ? this.results.filter(this.filterFn, this) : [];
+        const searchResults: OBJECT[] = this.results ? this.results.filter(this.filterFn, this) : [];
 
-        if (searchResult.length < this.results?.length) {
-            this.notifyLoadedData(searchResult);
+        if (searchResults.length !== this.filteredResults?.length) {
```
ashklianko, line 101: That approach isn't working if you first filter data with one search string that returns 3 items, and then paste into the filter input other string that is supposed to return other 3 items - search won't refresh

ashklianko (changes requested):

ashklianko (approved):

Issue #2652, opened 2022-06-27
Modal dialogs: comboboxes load all items after expanding


ashklianko:
@ashklianko commented on [Mon Jun 20 2022](https://github.com/enonic/app-contentstudio/issues/4708)

Open any Issues or Tasks dialog and expand Assignees/Items combobox: all items get loaded immediately in spite of they are supposed to be lazily loaded

PR #2653, opened 2022-06-27
Modal dialogs: comboboxes load all items after expanding #2652

ashklianko:
-Start loading combobox further only when scrolled down to the bottom


ashklianko:
@sgauruseu this is potentially may break some stuff, please pay attention on this task

alansemenov (approved):

PR #2708, opened 2022-08-08
Single-selection option-set does not load option form #2704

ashklianko:


ashklianko:
Asked @sgauruseu to check failed studio tests with this branch locally - all passed, thus merging now to let tests that failed in studio be fixed

PR #2739, opened 2022-09-01
Optimise performance of option container inside an option-set

alansemenov:
Layout option container only when the option is selected in an option-set


ashklianko (changes requested):
1. Open saved OptionSet content. Select option with inputs to be rendered after selection, and then deselect it: Save button remains ON 
2.  Reset button is not aligned with More button icon:
image

ashklianko (approved):

PR #2781, opened 2022-09-28
Support right-to-left directionality for text inputs and components #…

ashklianko:
…2771


src/main/resources/assets/admin/common/js/locale/Locale.ts
```diff
@@ -21,6 +22,14 @@ export class Locale
     private country: string;
     private displayCountry: string;
 
+    public static isRtlCode(code: string): boolean {
```
alansemenov, line 25: Should be called "`supportsRtl`"

src/main/resources/assets/admin/common/js/form/inputtype/text/TextInputType.ts
```diff
@@ -50,37 +51,43 @@ export abstract class TextInputType
...
 
+    protected updateInputLangParams(inputEl: FormInputEl): void {
+        const locale: string = this.getContext().formContext.getLanguage();
+
+        if (!StringHelper.isEmpty(locale)) {
+            const language: string = Locale.extractLanguage(locale);
+            inputEl.setLang(language);
+
+            if (Locale.isRtlCode(language)) {
+                inputEl.getEl().setAttribute('dir', 'rtl'); // doesn't work since reset.less in enonic-admin-artifacts...
```
alansemenov, line 85: Let's implement `setDir`, similar to `setLang`. `.style.direction` can be set there as well.

alansemenov (changes requested):

ashklianko:
Updated

src/main/resources/assets/admin/common/js/dom/Element.ts
```diff
@@ -531,6 +531,12 @@ export class Element {
         return this.getEl().getAttribute('lang');
     }
 
+    setDir(value: string): Element {
+        this.getEl().setAttribute('dir', value); // doesn't work since reset.less in enonic-admin-artifacts...
+        this.getHTMLElement().style.direction = value; // ...sets direction style to ltr, so overwriting style
+        return this;
```
alansemenov, line 537: Have released `enonic-admin-artifacts` 1.8.3 which no longer overrides `directory`. Let's upgrade to that (in both lib-admin-ui and CS) and remove the `style` override (and the comments).

alansemenov (changes requested):

alansemenov (approved):

Issue #2786, opened 2022-10-03
Method `highglightItemById` should have control on `notifyHighlightChanged`


reisfmb:
This method will be used in schemas visualization, and there's the need to avoid notifying highlight changes in some specific cases.

PR #2787, opened 2022-10-03
Added notify parameter to highlightItemById (#2786)

reisfmb:


src/main/resources/assets/admin/common/js/ui/treegrid/TreeGrid.ts
```diff
@@ -1698,7 +1698,7 @@ export class TreeGrid<DATA extends IDentifiable>
         this.setActive(true);
     }
 
-    private highlightRowByNode(node: TreeNode<DATA>) {
+    private highlightRowByNode(node: TreeNode<DATA>, notify: boolean = true) {
```
alansemenov, line 1701: We usually do it the other way around: call such a parameter `silent` and make it false by default. You will find this all over the code.

src/main/resources/assets/admin/common/js/ui/treegrid/TreeGrid.ts
```diff
@@ -1898,7 +1901,7 @@ export class TreeGrid<DATA extends IDentifiable>
         return nodes.map((node: TreeNode<DATA>) => node.getData());
     }
 
-    highlightItemById(dataId: string, expand: boolean = false) {
+    highlightItemById(dataId: string, notify: boolean = true, expand: boolean = false) {
```
alansemenov, line 1904: If you add a new argument to an existing public method you MUST make it last, otherwise it might break calls of this method.

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/ui/treegrid/TreeGrid.ts
reisfmb: Yep, I'm aware of this, but this is a new method Sasha implemented, so I believe no other place uses it so far...

PR #2794, opened 2022-10-05
Multiple occurrences of an option-set should not be auto-expanded

alansemenov:
Don't expand selected option in a collapsed option-set.


src/main/resources/assets/admin/common/js/form/set/optionset/FormOptionSetOccurrenceViewSingleOption.ts
```diff
@@ -146,16 +146,16 @@ export class FormOptionSetOccurrenceViewSingleOption
...
-            this.singleSelectionDropdown.hasSelectedOption() &&
-            optionView.getFormItemViews().length > 0;
-
-        this.setContainerVisible(shouldBeVisible);
-
+        this.setContainerVisible(this.isOptionViewShouldBeVisible(optionView));
         super.handleSelectionChanged(optionView);
     }
 
+    private isOptionViewShouldBeVisible(optionView: FormOptionSetOptionView): boolean {
```
alansemenov, line 153: "`is should`" won't work. Let's rename to "`isContainerExpansionRequired`"

PR #2799, opened 2022-10-06
Preselection of options inside option-set doesn't work in site.xml #2797

ashklianko:
-Populating property array of a FormOptionSetOptionView with a default selected options on add

-Adjusted constructor props of ApplicationConfiguratorDialog

### DIFF
--- a/src/main/resources/assets/admin/common/js/form/set/optionset/FormOptionSetOptionView.ts
+++ b/src/main/resources/assets/admin/common/js/form/set/optionset/FormOptionSetOptionView.ts
@@ -103,7 +103,14 @@ export class FormOptionSetOptionView
             this.appendChild(this.optionItemsContainer);
         }
 
+        const isDefaultSelected: boolean =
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/admin/common/js/form/set/optionset/FormOptionSetOptionView.ts
```diff
@@ -103,8 +103,15 @@ export class FormOptionSetOptionView
             this.appendChild(this.optionItemsContainer);
         }
 
+        const isDefaultSelected: boolean =
```
alansemenov, line 106: "`isDefaultSelected`" is misleading. It's not selected yet. We do a check to find out if it should be selected, then call `setSelected()` if it should. The most fitting is "`isDefaultOption`" (as it's declared as default in the xml schema).

alansemenov: Yeah, let's go for `isDefaultAndNew`

alansemenov (approved):

PR #2809, opened 2022-10-14
Preselection of options inside option-set doesn't work in site.xml #2797

ashklianko:
-Populating property array of a FormOptionSetOptionView with a default selected options on add -Adjusted constructor props of ApplicationConfiguratorDialog -Removed redundant code that was selecting default option -Added new param "isNew" to FormItemLayer indicating that a FormItem is newly added (can be item in just created content, can be added via user action, can be item in the app configuration form or in inspection panels)


alansemenov (approved):
Yep, this was a good change

Issue #2819, opened 2022-10-20
Incorrect placement of the reset link in the filter panel


alansemenov:
@sgauruseu commented on [Wed Apr 17 2019](https://github.com/enonic/app-contentstudio/issues/439)

**Filter Panel:**

![image](https://user-images.githubusercontent.com/3728712/56269949-ad2e5d00-60fd-11e9-8559-8ca8a8b51dc0.png)

PR #2821, opened 2022-10-20
Converted filter panel reset link into an icon (#2819)

alansemenov:


ashklianko:
On large screens close icon is here:
image

and on smaller screens it is here:
image

Is it intended or should the close icon remain on same position?

alansemenov:
> Is it intended or should the close icon remain on same position?

Yes, it should always be right-aligned relative to the entire panel. It's not only the search input this icon resets, but the entire panel.

ashklianko (approved):

Issue #2873, opened 2022-11-17
Inputs with the same name from different mixins break option/item set


alansemenov:
Conditions for an error: an item-set or an option-set that uses several different mixins containing an input with the same name.

The schema below is using two mixins, both of which have an input called "format". Resolution of the selected option form's title is trying to find an input by its name and fails because there are two.

Resolve form input by path instead.

```xml
<option-set name="type">
  <label i18n="schema.config.form.type.optionSet">Input type</label>
  <occurrences minimum="1" maximum="1"/>
  <options minimum="1" maximum="1">
    <option name="boolean">
      <label>Boolean</label>
      <items>
        <mixin name="schema-boolean-formats"/>
      </items>
    </option>
    <option name="string">
      <label>String</label>
      <items>
        <mixin name="schema-string-formats"/>
      </items>
    </option>
  </options>
</option-set>
```

**schema-boolean-formats:**
```xml
<mixin>
  <display-name>JSONSchema Boolean formats</display-name>
  <form>
    <input name="format" type="RadioButton">
      <label>Format</label>
      <config>
        <option value="default">Default (none)</option>
        <option value="consent">Consent</option>
      </config>
      <default>consent</default>
    </input>
  </form>
</mixin>
```
**schema-string-formats:**
```xml
<mixin>
    <display-name>JSONSchema String Formats</display-name>
    <form>
        <input name="format" type="RadioButton">
            <label>Format</label>
            <config>
                <option value="default" i18n="schema.string.format.default">Default (none)</option>
                <option value="zipcode" i18n="schema.string.format.zip">Zip</option>
            </config>
            <default>default</default>
        </input>
    </form>
</mixin>
```

PR #2874, opened 2022-11-17
Inputs with the same name from different mixins break option/item set (#2873)

alansemenov:


ashklianko (changes requested):
Works for the example in the task description, however fails when content type with such config present:
```
<content-type>
  <display-name>freeform</display-name>
  <super-type>base:structured</super-type>
  <allow-child-content>true</allow-child-content>
  <form>
    <mixin name="schema-boolean-formats"/>
    <mixin name="schema-string-formats"/>
  </form>
</content-type>
```

alansemenov:
> Works for the example in the task description, however fails when content type with such config present:
> 
> ```
> <content-type>
>   <display-name>freeform</display-name>
>   <super-type>base:structured</super-type>
>   <allow-child-content>true</allow-child-content>
>   <form>
>     <mixin name="schema-boolean-formats"/>
>     <mixin name="schema-string-formats"/>
>   </form>
> </content-type>
> ```

This is misuse of the mixins - same as if you would put two inputs with the same name next to each other. So it doesn't have anything to do with this bug, which is when you have inputs with the same name under different parents (which is allowed).

ashklianko (approved):

Issue #2876, opened 2022-11-18
Focus is incorrectly set after resetting an option in Option-set


alansemenov:
1. Create a content with an option-set.
2. Select any option.
3. Save
4. Open context-menu for the selected option and select "Reset"

**Bug 1**: Focus is not set inside the dropdown's input field.

5. Press the "down" button
6. Dropdown expands and instantly disappears

**Bug 2**: It's no longer possible to set focus inside the input field.

PR #2877, opened 2022-11-18
Focus is incorrectly set after resetting an option in Option-set #2876

ashklianko:


alansemenov (changes requested):
The second bug is fixed, but focus is still not set on the dropdown's input field. After "Reset" you should be able to instantly start typing to find another option.

Also, if you press "Esc" when the dropdown is expanded, the input will loose the focus.

alansemenov (approved):

Issue #2920, opened 2022-12-15
Optimize build time


edloidas:
Update the build scripts.

edloidas:
Using `--cache` for ESLint decreases linting time:
| Type | Time |
| -- | -- |
| No cache | ~14.5s |
| Cache w/o changes | ~2.2s |
| Cache w/ changes | ~4.75s |

On Windows, command execution time can be measured with `Measure-Command { npm run lint:quiet }`

edloidas:
Another step to optimize the build is to use `swc` to generate JS during dev build.
`swc` doesn't care about types, so in case of typescript we must check types separately. It can be done with `--noEmit` flag to make it a little bit faster, or with a full types generation. Because of the flags like `--skipLibCheck` usage, we might not be 100% the types are correct, so I suggest that we use a regular `tsc` for types generation in production mode, that will run on CI.

| Type | Time |
| -- | -- |
| `tsc` dev | ~15.1s |
| `tsc` prod | ~14.7s |
| `swc` + `tsc` types | ~10s |
| `swc` + `tsc` typecheck, `--noEmit` | ~9.5s |

edloidas:
Dropping the support of the ES5 also adds some minor improvement (that is not actually visible, because `swc` is fast enough). But, what is more importantly the resulted non-minified JS becomes smaller ~20%.

edloidas:
Migrating to ES6 must be done in a separate task, as it appears, that:
* Apps, that are using `lib-admin-ui` `lib` part must also be compiled into ES6, otherwise it will lead to:
  `Class constructor cannot be invoked without new` exceptions
* Some `lib-admin-ui` libraries, like `slickgrid`, rely on ES5 and can not be used with ES6+, e.g. [function regex](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/lib/slickgrid/slick.dataview.js#L682)  does not support arrow functions, there.

edloidas:
Enforcing rules of Enonic's ESLint [config](https://github.com/enonic/enonic-npm-modules/tree/master/packages/eslint-config) will make typecheck faster simply by making developers to avoid implicit `any`, function's implicit return types, and other things, that make compiler to do it itself on each compilation.

edloidas:
[...]

PR #2931, opened 2022-12-20
Optimize build time #2920 

edloidas:
* Added ESLint caching.
* Updated ESLint to use Enonic config.
* Optimized TSConfig to a single file.
* Split dev lib build into two tasks: first will use swc to compile JS, second will use tsc to create types.
* Added new gradle task to copy compiled sourced in dev mode, because swc can not be configured to do it itself.
* Added SWC loader for ts files in Webpack.
* Added ESLint rule to track missing super() calls.
* Fixed missing super() calls in constructors.
* Removed unused lint npm script.
* Made Gradle to run a single npm script for optimized performance, since running two npm tasks will take more time.
* Reorganized build scripts.

_All measurements and other notes were added to comments inside the task._


alansemenov (approved):
Great job!

Issue #2949, opened 2022-12-29
ContentTypeSummary: add allowChildContentType field


ashklianko:
ContentTypeSummary has a field that we're not returning from the backend: `allowChildContentType ` 
This field might be useful in certain scenarios

PR #2950, opened 2022-12-29
ContentTypeSummary: add allowChildContentType field #2949

ashklianko:


src/main/resources/assets/admin/common/js/schema/content/ContentTypeSummary.ts
```diff
@@ -10,26 +10,29 @@ export class ContentTypeSummary
...
-    private modifier: string;
+    private readonly modifier: string;
 
-    private owner: string;
+    private readonly owner: string;
 
-    private metadata: MixinNames;
+    private readonly metadata: MixinNames;
+
+    private readonly allowChildContentType: string[];
```
alansemenov, line 31: We should rename this to `allowedChildContentTypes` (I know it's called like that in API, but we don't have to follow incorrect naming in TS classes).

src/main/resources/assets/admin/common/js/schema/content/ContentTypeSummary.ts
```diff
@@ -188,8 +196,11 @@ export class ContentTypeSummaryBuilder
 
     metadata: MixinNames;
 
+    allowChildContentType: string[];
```
alansemenov, line 199: This should be renamed as well.

src/main/resources/assets/admin/common/js/schema/content/ContentTypeSummary.ts
```diff
@@ -118,6 +122,10 @@ export class ContentTypeSummary
         return this.metadata;
     }
 
+    getAllowChildContentType(): string[] {
+        return this.allowChildContentType;
+    }
```
alansemenov, line 127: Instead of this we need `isMediaChildContentAllowed`:

```Typescript
isMediaChildContentAllowed(): boolean {
  return this.allowChildContent &&
      this.allowedChildContentTypes?.some((name: string) => new ContentTypeName(name).isDescendantOfMedia());
}
```

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/schema/content/ContentTypeSummary.ts
ashklianko: Disagree with this part: let's keep ContentTypeSummary.ts as a data class without non-related logic. Child check is moved to ContentTypesHelper.ts in studio

Issue #2918, opened 2022-12-14
Option-set with only one option set to default is not expanded


poi33:
![image](https://user-images.githubusercontent.com/3340349/207586128-6dd98625-79e1-48fb-96b2-9fee1e6c4493.png)

Selecting it and unselecting it does expand it:
![image](https://user-images.githubusercontent.com/3340349/207586206-4ac190c0-da3a-4750-b167-6e634914a7fe.png)

```
Layout mixin
?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<layout>
    <display-name>1 column layout</display-name>
    <form>
        <option-set name="overlap">
            <label>Overlapping layout</label>
            <options minimum="0" maximum="1">
                <option name="layout">
                    <default>true</default>
                    <label>Color of overlapping part</label>
                    <items>
                        <mixin name="theme-color-picker"/>
                    </items>
                </option>
            </options>
        </option-set>
    </form>
    <regions>
        <region name="main"/>
    </regions>
</layout>

    
    
    // mixin xml
    <?xml version="1.0" encoding="UTF-8"?>
<mixin>
    <display-name>Theme color picker</display-name>
    <form>
        <input name="themeColor" type="RadioButton">
            <label>Theme color</label>
            <occurrences minimum="0" maximum="1"/>
            <config>
              <option value="white">White</option>
              <option value="grey">Grey</option>
              <option value="blue">Blue</option>
              <option value="black">Black</option>
            </config>
        </input>
    </form>
</mixin>

```

alansemenov:
@ashklianko check if this is still relevant after the latest fixes

PR #2955, opened 2023-01-02
Option-set with only one option set to default is not expanded #2918

ashklianko:


ashklianko:
Tests are OK: 
https://github.com/enonic/app-contentstudio/pull/5650

alansemenov (approved):

PR #3067, opened 2023-04-05
Wrapper element inside a region breaks the Page Editor

alansemenov:


src/main/resources/assets/admin/common/js/dom/Element.ts
```diff
@@ -1336,9 +1339,12 @@ export class Element {
     }
 
     private unregisterChildElement(child: Element): number {
```
ashklianko, line 1341: Now sense in returning index now

src/main/resources/assets/admin/common/js/dom/Element.ts
```diff
@@ -1336,9 +1339,12 @@ export class Element {
...
-        let childIndex = this.children.indexOf(child);
+        if (!this.children?.length || !child.parentElement) {
+            return -1;
+        }
+        const childIndex = this.children.indexOf(child);
         if (childIndex < 0) {
-            throw new Error('Child element to remove not found');
+            this.children.forEach((c: Element) => c.unregisterChildElement(child));
         }
         this.children.splice(childIndex, 1);
```
ashklianko, line 1349: this will be invoked at each potential parent where child is being looked for

src/main/resources/assets/admin/common/js/dom/Element.ts
```diff
@@ -1336,9 +1339,12 @@ export class Element {
...
         child.parentElement = null;
```
ashklianko, line 1350: this will be invoked at each potential parent where child is being looked for

ashklianko (commented):
`removeChildElement()` method will break since `child.notifyRemoved(this);` might be invoked with wrong parent since parent can be not equal to `this`

src/main/resources/assets/admin/common/js/dom/Element.ts
ashklianko, line 1349: this will be invoked at each potential parent where child is being looked for

ashklianko (changes requested):

src/main/resources/assets/admin/common/js/dom/Element.ts
ashklianko: This comment is addressed to 1349-1350 lines

src/main/resources/assets/admin/common/js/dom/Element.ts
alansemenov: Returned index is used [here](https://github.com/enonic/lib-admin-ui/blob/issue-3066/src/main/resources/assets/admin/common/js/dom/Element.ts#L694-L697).

ashklianko (changes requested):
1) `replaceWith()` will not remove `this` is not removed from parent's children
2) Need to set correct parent when notifying about removing child :
```
private removeChildElement(child: Element): Element {
        assertNotNull(child, 'Child element to insert cannot be null');

        const parent: Element = this.unregisterChildElement(child);

        child.notifyRemoved(parent);
        return this;
    }
```

ashklianko (approved):

Issue #3125, opened 2023-05-11
Validation for the Insert Link dialog's URL field is too strict


alansemenov:
These URLs won't be allowed:

https://oslo-universitetssykehus.no/seksjon/regionalt-senter-for-astma-allergi-og-overfolsomhet/Documents/Matprovokasjoner%20-%20servering%20og%20h%C3%A5ndtering%20av%20allergener%20Versjon%202,%202021.pdf

https://example.sharepoint.com/sites/exampledepartment/Shared%20Documents/Forms/AllItems.as[…]Personal%2Fvalidplease&viewid=3204ecd9-3360-499d-b8

PR #3126, opened 2023-05-19
Allow commas in path section of url #3125

reisfmb:
As requested, url path section now allow "," and "…"


alansemenov (changes requested):
There are two cases described in the task. This fix is only for the first one, we need to fix the other case as well.

reisfmb:
> There are two cases described in the task. This fix is only for the first one, we need to fix the other case as well.

I didn't fix the other case because it didn't raised any errors when I inserted it on insert link dialog.

alansemenov:
@reisfmb it fails for me instantly because of `[...]` in the url

image

alansemenov (approved):

PR #3259, opened 2023-09-18
Grid navigation improvements

alansemenov:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=4.8.0-SNAPSHOT
+version=4.8.0-3258-SNAPSHOT
```
ashklianko, line 1: Need to set the correct version

ashklianko (changes requested):

ashklianko (approved):

Issue #3335, opened 2023-11-02
OptionsTreeGrid: dead code removal


ashklianko:
Found some code that is never used

PR #3336, opened 2023-11-02
OptionsTreeGrid: dead code removal #3335

ashklianko:


ashklianko:
Checked apps that rely on lib admin as studio, app-main, users, applications and studio plus, no trails of usage

alansemenov (approved):

Issue #3356, opened 2023-11-20
LazyListBox: items are not loaded, if the last is invisible


edloidas:
If the last child is invisible, the intersection observer won't trigger any events. The last visible (in DOM) item must be used instead in the observer.

PR #3357, opened 2023-11-20
LazyListBox: items are not loaded, if the last is invisible #3356

edloidas:


edloidas:
@ashklianko When they are hidden, obviously. Let's say you have a list of elements and a toggler that would trigger a simple filter, and hide (using CSS) the elements from view. If the last added item is not visible, lazy loading will unfortunately not work.
LizyListBox doesn't have complex logic like Grid and renders all elements into the DOM, so it takes a long time to re-render a large list, and removing and adding specific elements from the list only complicates the logic and is slower than CSS anyway.

alansemenov (approved):

src/main/resources/assets/admin/common/js/ui/selector/list/LazyListBox.ts
```diff
@@ -43,35 +43,40 @@ export class LazyListBox<T> extends ListBox<T> {
...
-        this.addLazyLoadWhenLastIsVisible(this.getLastChild());
+        this.addLazyLoadWhenLastIsVisible(lastVisibleChild);
+    }
+
+    protected getLastVisibleChild(): Element | undefined {
+        if (!this.isVisible() || this.getItemCount() === 0) {
+            return undefined;
+        }
+
+        return [...this.getChildren()].reverse().find((item: Element) => {
```
alansemenov, line 77:
```suggestion
        return [...this.getChildren()].reverse().find((item: Element) => item.isVisible());
```

alansemenov (approved):

PR #3489, opened 2024-03-07
Set data-path attribute on inputs #3488

### LINKED ISSUE #3488: Set data-path attribute on inputs
AI assistant needs a way to find editable items in DOM, need  to add a data-attribute with path so Assistant could find it via selector.

ashklianko:
Input element's path is a dynamic thing, it changes when occurrences are moved/added/removed, so it is only valid when someone asks for it in a particular moment. However this path is required by the AI Assistant to be able to find the input with selector query, thus need to update the data-path attribute constantly to keep it actual

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ src/main/resources/assets/spec/coverage/
 
 # Visual Studio Code
 /bin/
+.vscode/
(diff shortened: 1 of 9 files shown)


alansemenov (changes requested):
We need to rename `isSagaEditableType` and `isSagaEditable` to something more generic, like `isManagedType`/`isManaged`

PR #3525, opened 2024-04-05
Non-selectable options in dropdowns should not be interactable #3521

### LINKED ISSUE #3521: Non-selectable options in dropdowns should not be interactable
Checkboxes on non-selectable options inside dropdowns should be styled as disabled. Now they look exactly like selectable ones - they can be clicked and unselected, but this will not affect selection when Apply is pressed. To avoid confusion they should not be interactable.

Expected look (the top 3 ones and the bottom one are non-selectable):
image

ashklianko:
-Making rows be marked as 'readonly' rather than cells


ashklianko:
The branch is ok and can be checked, however marked as draft because there are potential methods to remove:
`Dropdown.markReadOnly` is only used in CompareContentVersionsDialog, and the latest can be updated to set readonly param on option creation

PR #3527, opened 2024-04-08
Non-selectable options in dropdowns should not be interactable

ashklianko:


ashklianko:
Requires https://github.com/enonic/app-contentstudio/pull/7495 to not make ContentStudio build fail

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=4.9.0-SNAPSHOT
+version=4.9.0-3521-SNAPSHOT
```
alansemenov, line 1: Remove the temporary version please

alansemenov (changes requested):
Remove the temporary version please

alansemenov (approved):

PR #3618, opened 2024-06-25
Add level navigation for item sets

ashklianko:
#332


alansemenov (commented):
It's not directly related to this PR, but I suggest we rename all the "`Saga*`" classes to be more name-agnostic, f.ex. `SagaHelper` -> `AIHelper` etc.

edloidas (approved):

PR #3723, opened 2024-09-24
Remove SlickGrid

ashklianko:


ashklianko:
Just need to update build.gradle for correct version, current version is used for the studio tests to run

ashklianko:
Need to update `gradle.properties` fille to set correct app version

alansemenov (approved):

PR #3785, opened 2024-11-13
Don't hide the Juke icon when focus is inside the input #3784

ashklianko:


edloidas:
This is solved in https://github.com/enonic/lib-admin-ui/pull/3786 , because Ai controls were remade there and we have different styles and structure now. I suggest we just close this PR, if we merrge the animations.

Issue #3802, opened 2024-11-26
Set active element for context 


ashklianko:
Required for https://github.com/enonic/app-contentstudio/issues/8089
Adjust AIHelper and related code by adding a method for setting active button and add required styles for AI toggle button

PR #3803, opened 2024-11-27
Set active element for context #3802

ashklianko:


ashklianko:
Required for https://github.com/enonic/app-contentstudio/pull/8102 and https://github.com/enonic/app-ai-content-operator/pull/150

src/main/resources/assets/admin/common/styles/api/ai/ai-dialog-control.less
```diff
@@ -10,8 +10,15 @@ button.ai-dialog-control {
...
+  filter: grayscale(0.50);
+  opacity: 0.5;
 
-  &:not(:hover):not(:focus):not(:focus-within):not(.input-focused) {
+  &:hover, &.active {
+    filter: none;
+    opacity: 1;
+  }
+  
+  &:not(:hover):not(:focus):not(:focus-within):not(.input-focused):not(.active) {
```
edloidas, line 21: Can drop `:focus`, since `:focus-within` will work the same for root element.

src/main/resources/assets/admin/common/js/ai/AiHelper.ts
```diff
@@ -74,17 +79,26 @@ export class AiHelper {
...
-        AiHelper.DIALOG_CONTROL_REGISTRY.set(this.config.aiButtonContainer, aiIcon);
+        AiHelper.getOrCreateIconsRegistry().set(this.config.aiButtonContainer, aiIcon);
 
         return aiIcon;
     }
 
+    private static getOrCreateIconsRegistry(): Map<Element, AiDialogControl> {
+        if (!Store.instance().has(AI_ICONS_REGISTRY_KEY)) {
+            console.log('Creating new AI icons registry');
+            Store.instance().set(AI_ICONS_REGISTRY_KEY, new Map());
```
edloidas, line 96: We need to rewrite this and use `WeakMap` here, since keys are objects. The thing is, objects won't be release and garbage collected, even if there are no reference to them left.
The bad part: we won't be able to use `forEach` on WeakMap. This, instead of iteration, we better switch to event-based updates. But must be careful with the implementation: events could fire/handled in different order, so I suggest events to signal only state update, while the actual state will be stored here and accessed in AiDialogControl from single place (we can store it here, in AiHelper).

edloidas (changes requested):

edloidas (approved):

Issue #3813, opened 2024-12-06
Make AI configuration for inputs more customizable


ashklianko:
We utilize the isAiEditable(): boolean method to identify inputs intended for AI injection. However, this approach is insufficient for comprehensive configuration, as it restricts us to a single parameter.

Certain inputs require both the AI icon to open the AI dialog and a translation status icon, while others may need only one of these features.

To address this limitation, introduce an enhanced AI configuration for inputs. It should enable the specification of any combination of AI tools and include a context parameter to distinguish between different inputs sets, such as  data, pages, or x-data.

PR #3814, opened 2024-12-06
Make AI configuration for inputs more customizable #3813

ashklianko:


ashklianko:
Required for https://github.com/enonic/app-contentstudio/pull/8140

alansemenov (approved):

src/main/resources/assets/admin/common/js/ai/AiTool.ts
```diff
@@ -0,0 +1,10 @@
+
+export enum AiTool {
+    OPEN_AI_DIALOG,
+    AI_STATE,
+}
```
edloidas, line 5: I'd suggest adding string values to enum entries. It would better for debugging to see string values somewhere down the flow, instead of just `0`.
Otherwise, it will be compiled into:
```js
(function (AiTool) {
    AiTool[AiTool["OPEN_AI_DIALOG"] = 0] = "OPEN_AI_DIALOG";
    AiTool[AiTool["AI_STATE"] = 1] = "AI_STATE";
})(AiTool || (AiTool = {}));
```
instead of 
```js
(function (AiTool) {
    AiTool["OPEN_AI_DIALOG"] = "openAiDialog";
    AiTool["AI_STATE"] = "aiState";
})(AiTool || (AiTool = {}));
```

src/main/resources/assets/admin/common/js/form/FormContext.ts
```diff
@@ -14,14 +17,15 @@ export class FormContext {
 
     private validationErrors: ValidationError[];
 
-    private readonly aiEditable: boolean;
+    private readonly aiTools: AiTool[];
```
edloidas, line 20: Think it's time we start using `Set`. It will result in more readable and simple code in our case, since we add unique values and try to find them later with `Array.some`

edloidas (approved):

PR #3850, opened 2025-01-21
Remove usage of IsAuthenticatedRequest and LoginResult #3849

ashklianko:


.github/workflows/gradle.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     outputs:
-      publish: ${{ steps.publish_vars.outputs.release != 'true' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/4.')) }}
+      publish: ${{ steps.publish_vars.outputs.release != 'true' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/issue-3849') }}
```
rymsha, line 13: This checks are written to prevent unwanted releases. Now we have a new "release" that will be picked as latest-gratest

rymsha (changes requested):

.github/workflows/gradle.yml
alansemenov: this is temporary, we'll remove this before merging.

alansemenov (approved):
Looking good. Don't forget to remove temporary changes in gradle.yml and gradle.properties.

rymsha (approved):

Issue #3867, opened 2024-10-17
Wrong element is triggered after pressing enter key in a selector


sgauruseu:
1. Open Insert Image dialog
2. Move the cursor to `Options Filter` input
3. Press `Enter` key

**BUG**  - Tree mode is switched on after pressing `Enter` key

![image](https://github.com/user-attachments/assets/4adfc4ab-60d8-48a3-bd67-303499263ed9)

ashklianko:
Same thing happens in our Link dialog, but upload button is the one getting 'pressed' instead of the selector's input.

Issue was that `button` elements around selector, like a mode toggler or an uploader, don't have a type attribute set: [Always specify the type attribute for the <button> element. Different browsers may use different default types for the <button> element.](https://www.w3schools.com/Tags/att_button_type.asp)

PR #3868, opened 2025-02-05
Wrong element is triggered after pressing enter key in a selector #3867

ashklianko:


ashklianko:
This one should be both in 5.0 and master

alansemenov (approved):

Issue #3880, opened 2025-02-13
TreeListBox: don't invoke lazy load on shown


ashklianko:
When to load items in the list should be up to those who use this list

ashklianko:
This change is required to make lists reload on show on demand

PR #3881, opened 2025-02-13
TreeListBox: don't invoke lazy load on shown #3880

ashklianko:
TreeListBox: don't invoke lazy load on shown #3880

- Trigger load on first show when it is needed


ashklianko:
Made CS to run tests on this branch: version=5.0.0-SNAPSHOT

alansemenov (approved):

PR #3900, opened 2025-02-25
Incorrect scrolling behaviour in the grid #3899

ashklianko:


alansemenov:
It works correctly most of the time. But in some cases (I found one) it scrolls too much when I press down arrow just once. See the attached video. I reproduced it several times on this specific node.

https://github.com/user-attachments/assets/d17e944b-4ddb-4bcf-a0d4-e1d49acf4690

alansemenov (approved):
good stuff 👍

Issue #3934, opened 2025-03-19
Inline padding is hardcoded in TreeListBox


ashklianko:
TreeListBox currently calculates padding for nested children so they look like a proper tree, however this approach is rigid and hard to override for a custom list.

Use html/css/less features instead.

NB: changes will require to update CS, CS+, Users and Applications

PR #3935, opened 2025-03-19
TreeListBox: get rid of hardcoded padding calculation #3934

ashklianko:


src/main/resources/assets/admin/common/js/ui/selector/list/TreeListBox.ts
```diff
@@ -128,20 +124,33 @@ export abstract class TreeListBox<I> extends LazyListBox<I> {
         return this.scrollParent;
     }
 
+    calcLevel(): number {
```
alansemenov, line 127: Let's rename to `getLevel()`

alansemenov (changes requested):

alansemenov (approved):

Issue #3938, opened 2025-03-19
Focusing an input will no longer set it in the Juke Operator's context


alansemenov:
When Juke Operator dialog is active/open and user clicks inside an input, the focused input should be set in the dialog's context. That was one of the main reasons why we didn't make the dialog modal. Basically focusing an input should do the same as clicking on the Juke icon next to it.

PR #3945, opened 2025-03-20
Focusing an input must set a context in Juke Operator #3938

ashklianko:


ashklianko:
Requires https://github.com/enonic/app-ai-content-operator/pull/341

alansemenov (changes requested):
It works fine but not for the Display Text input

alansemenov (approved):

PR #3961, opened 2025-03-26
Insufficient padding inside a set

alansemenov:


alansemenov:
Before (left) vs after (right):

Screenshot 2025-03-27 at 09 09 58
Screenshot 2025-03-27 at 09 12 15

ashklianko (approved):

PR #3972, opened 2025-03-31
Preview spinner is delayed #3971

pmi:


alansemenov (changes requested):
Nice improvement! But we need to improve it a bit more :)

Before this fix, when you quickly navigate from item 1 to item 5, Content Studio will only load preview for item 5. After this change, it will always load preview for item 2, then for item 5. We should get rid of this extra request. 

See the attached videos (tested on throttled speed, then it becomes very obvious):

Before the change:

https://github.com/user-attachments/assets/b9da84d1-7b43-4abd-8c84-0fb7b4e46e78

After the change:

https://github.com/user-attachments/assets/60893bb9-a76f-4e7e-9418-8b0b0537b53a

ashklianko:
Also plz see my comment https://github.com/enonic/lib-admin-ui/pull/3973#pullrequestreview-2731495690

alansemenov:
Yes, we definitely don't want any changes in debounce behaviour - those were carefully adjusted over the years. What we need is that spinner is shown the moment selection leaves the current item, the way it was before the Liveview changes.

alansemenov (approved):

src/main/resources/assets/admin/common/js/app/browse/BrowsePanel.ts
```diff
@@ -42,7 +44,7 @@ export class BrowsePanel
 
     protected toggleFilterPanelButton: ActionButton;
 
-    private debouncedActionsAndPreviewUpdate: () => void;
+    protected keyNavigator: SelectableListBoxKeyNavigator<any>;
```
alansemenov, line 47: We shouldn't use `<any>`

src/main/resources/assets/admin/common/js/app/browse/BrowsePanel.ts
```diff
@@ -104,26 +105,65 @@ export class BrowsePanel
...
+                // mark recently selected content before selectionChangedListener reads it
+                // to prevent loading content when scrolling with keyboard
+                if (!recentlySelected) {
+                    recentlySelected = true;
+                }
+                return false;
+            }
+            this.keyNavigator.onKeyDown(upDownHandler);
+            this.keyNavigator.onKeyUp(upDownHandler);
+        }
```
alansemenov, line 147: Instead of using `keyNavigator` in the base `BrowsePanel` class we should change `initTreeGridListeners` to be protected and add the additional logic (where `keyNavigator` is used) in Content Studio's `ContentBrowsePanel`. `recentlySelected` will then have to be made a protected class property as well.

alansemenov (changes requested):

src/main/resources/assets/admin/common/js/app/browse/BrowsePanel.ts
alansemenov: Actually, `recentlySelected` should be private and there should be protected methods called `setRecentlySelected` and `resetRecentlySelected` in BrowsePanel

src/main/resources/assets/admin/common/js/app/browse/BrowsePanel.ts
```diff
@@ -104,26 +105,65 @@ export class BrowsePanel
...
         this.selectableListBoxPanel?.getToolbar().getSelectionPanelToggler().onActiveChanged(isActive => {
             this.toggleSelectionMode(isActive);
         });
     }
 
-    private handleTreeListSelectionChanged() {
-        const totalFullSelected: number = this.selectableListBoxPanel.getSelectedItems().length;
+    protected createKeyNavigator() {
+        return undefined;
+    }
```
alansemenov, line 156: Remove

src/main/resources/assets/admin/common/js/app/browse/BrowsePanel.ts
pmi: Well, having it in BrowsePanel expands that functionality to all descendants.

If you on the other hand make `initTreeGridListeners` protected and put the logic to the ContentBrowsePanel, we're going to only have it ContentBrowsePanel.

Since keyNavigator is of type `SelectableTreeListBoxKeyNavigator` that comes from lib-admin-ui and is not exclusive to content studio, it makes perfect sense to have it in the base class.

And finally, pattern with using protected methods implemented by descendants is widely used in those classes so I just don't see a reason we can't use it here ?

pmi:
```    
    protected updateFilterPanelOnSelectionChange() {
        // TO BE IMPLEMENTED BY INHERITORS
    }

    protected enableSelectionMode() {
        // implement in inheritors
    }

    protected disableSelectionMode() {
        // implement in inheritors
    }

    protected createToolbar(): Toolbar<ToolbarConfig> {
        throw Error('Must be implemented by inheritors');
    }

    protected createListBoxPanel(): SelectableListBoxPanel<ViewItem> {
        throw Error('Must be implemented by inheritors');
    }

    protected createBrowseItemPanel(): BrowseItemPanel {
        throw Error('Must be implemented by inheritors');
    }
```

alansemenov:
>Well, having it in BrowsePanel expands that functionality to all descendants.

Not really. In this implementation it's spread between the base class and the inheriting class, one cannot be used without another. Which doesn't really make sense. Either we have everything in BrowsePanel (both initialisation of keyNavigator and behaviour) - which will make sense but then we need to find a way to pass proper type to SelectableListBoxKeyNavigator - or we have everything where it's needed. The first approach is better of course, why would we not handle key navigation in other classes inheriting from BrowsePanel? Plus as you said, SelectableListBoxKeyNavigator is in lib-admin-ui, so everything should be there.

> And finally, pattern with using protected methods implemented by descendants is widely used in those classes so I just don't see a reason we can't use it here ?

Yes, as you can see with all the create methods from your example, they throw exception if not implemented. In your case the base method returns `undefined`, we never do that for methods that don't return `void`.

pmi:
> > Well, having it in BrowsePanel expands that functionality to all descendants.
> 
> Not really. In this implementation it's spread between the base class and the inheriting class, one cannot be used without another. Which doesn't really make sense. Either we have everything in BrowsePanel (both initialisation of keyNavigator and behaviour) - which will make sense but then we need to find a way to pass proper type to SelectableListBoxKeyNavigator - or we have everything where it's needed. The first approach is better of course, why would we not handle key navigation in other classes inheriting from BrowsePanel? Plus as you said, SelectableListBoxKeyNavigator is in lib-admin-ui, so everything should be there.
>

Fine, then we make BrowsePanel generic and pass the type by descendants 👍 
This will require changes in the users and apps apps though.
 
> > And finally, pattern with using protected methods implemented by descendants is widely used in those classes so I just don't see a reason we can't use it here ?
> 
> Yes, as you can see with all the create methods from your example, they throw exception if not implemented. In your case the base method returns `undefined`, we never do that for methods that don't return `void`.

Exactly 3 out of 6 throw exception, the other 3 don't 😉

PR #3973, opened 2025-03-31
Preview spinner is delayed #3971

pmi:


ashklianko (changes requested):
Though these changes formally do trigger the load animation, they also have a serious flaw.
Selection change debounce is needed to not trigger costly operations that calculate available permissions, enabled buttons etc. if selection didn't change then no extra requests are sent, however in this approach those requests are sent immediately. 
Here is an example: try selecting/deselecting a content fast via selection checkbox. Compare network section before:
Screenshot 2025-04-01 083330

and in this branch after:
Screenshot 2025-04-01 082355


I believe we should keep debounce selection change handler as it is but add additional selection change listener that will directly toggle loading spinner.

Issue #3991, opened 2025-04-08
Extend Widget by URL functions


anatol-sialitski:
Add functions:

- getFullUrl
- getFullIconUrl

PR #3992, opened 2025-04-08
Extend Widget by URL functions #3991

anatol-sialitski:


src/main/resources/assets/admin/common/js/content/Widget.ts
```diff
@@ -43,6 +44,14 @@ export class Widget<B extends WidgetBuilder = WidgetBuilder, C extends WidgetCon
...
     }
 
+    public getFullUrl(): string {
+        return `${CONFIG.getString('widgetApiUrl')}/${this.url}`;
+    }
+
+    public getFullIconUrl(): string {
+        return `${CONFIG.getString('widgetApiUrl')}${this.iconUrl}`;
+    }
+
```
pmi, line 54: First one uses `/` between url parts and the second one doesn't, is it by design ?

src/main/resources/assets/admin/common/js/content/Widget.ts
```diff
@@ -43,6 +44,18 @@ export class Widget<B extends WidgetBuilder = WidgetBuilder, C extends WidgetCon
...
+    public getFullUrl(): string {
+        return `${this.getBaseUrl()}/${this.url}`;
+    }
+
+    public getFullIconUrl(): string {
+        return `${this.getBaseUrl()}${this.iconUrl}`;
+    }
+
+    private getBaseUrl(): string {
+        return (CONFIG.getString('widgetApiUrl') || '').replace(/\/+$/, '');
```
rymsha, line 56: Since it is not possible to mount `/_/` on `/` there should not be necessary to remove trailing slash ever

rymsha (approved):

Issue #4017, opened 2025-05-06
Move permission-selector.less to CS


ashklianko:
This file is not used in lib-admin

PR #4018, opened 2025-05-06
Move permission-selector.less to CS #4017

ashklianko:


ashklianko:
To be merged together with https://github.com/enonic/app-contentstudio/pull/8738

PR #4040, opened 2025-05-23
Move access/permissions texts to CS #4039

ashklianko:


ashklianko:
To be merged together with https://github.com/enonic/app-contentstudio/pull/8738

alansemenov (approved):

Issue #4132, opened 2025-05-27
Keyboard arrows in modal dialogs behave like tabbing


vot-enonic:
![Image](https://github.com/user-attachments/assets/e1d2e5f0-f91d-4154-a3ef-170d90670904)

PR #4133, opened 2025-08-25
Keyboard arrows in modal dialogs behave like tabbing #4132

ashklianko:


ashklianko:
Arrows must not be responsible for switching the focused element in modal dialogs, TAB and SHIFT + TAB do it. Arrows are supposed to switch the focus within a UI element https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/

Issue #4134, opened 2025-08-29
Enonic UI: Use ID context provider


edloidas:
To prevent issues with duplicated IDs on a page across different Virtual DOMs (roots) in Preact, we must wrap components in ]LegacyElement] or any other component that overrides `LegacyElement.renderJsx()` function into `IdProvider` component from Enonic UI and pass unique static prefix there, if necessary.

PR #4135, opened 2025-08-29
Enonic UI: Use ID context provider #4134

edloidas:
Upgraded Enonic UI version.
Moved all UI2 components to TSX.
Added ID context provider wrapper for elements.

### DIFF
--- a/src/main/resources/assets/admin/common/js/ui2/ActionIcon.tsx
+++ b/src/main/resources/assets/admin/common/js/ui2/ActionIcon.tsx
@@ -1,6 +1,5 @@
 import * as UI from '@enonic/ui';
 import {unwrap} from '@enonic/ui';
-import {nanoid} from 'nanoid';
 import {render} from 'react-dom';
 import {BrowserHelper} from '../BrowserHelper';
 import {Action} from '../ui/Action';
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/admin/common/js/ui2/ActionIcon.tsx
```diff
@@ -56,13 +56,19 @@ export class ActionIcon extends LegacyElement<typeof UI.IconButton> {
...
+        const ActionIconComponent = this.component;
+
+        render(
+            <UI.IdProvider prefix={prefix}>
+                <UI.Tooltip value={unwrap(props.title)}>
+                    <ActionIconComponent {...props} />
+                </UI.Tooltip>
+            </UI.IdProvider>,
+            this.getHTMLElement()
+        );
```
alansemenov, line 71: This method is virtually identical to the one in `ActionButton`, except for the component itself. And I assume it will be the same for other components as well. Shouldn't we implement a common helper method which wraps any component with an IdProvider, with optional tooltip? I see we already have one in `LegacyElement`, why not use that one?

edloidas: I've moved the prefix function to `LegacyComponent`. 

As for the rest of the code, I don't think it should be moved elsewhere for several reasons:
1. `renderJsx()` is essentially rendering the root and is meant to be overridden in scenarios where we need to insert additional components between the actual component and the provider.
2. This `render()` function is expected to include some duplication across different components. If we ever reach a point where it can be cleanly reused, we should extract it into a separate component. For now, though, the tooltip behavior is only needed in two places. Not that we in need of `TooltipWithIdProvider` component.
3. We should avoid inheritance whenever we're dealing with render logic. Otherwise, it creates unnecessary coupling and fragile dependencies between how elements are displayed.

Issue #4217, opened 2025-12-15
Implement Instant type


anatol-sialitski:
Implement Instant type

PR #4230, opened 2026-01-05
Implement Instant type #4217

anatol-sialitski:


src/main/resources/assets/admin/common/js/data/ValueTypeInstant.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        return this.isValid(value);
+    }
+
+    newValue(value: string): Value {
+        if (!value || !this.isConvertible(value)) {
+            return this.newNullValue();
+        }
+        let date: Instant = Instant.fromString(value);
```
alansemenov:
```suggestion
        const date: Instant = Instant.fromString(value);
```

src/main/resources/assets/admin/common/js/data/ValueTypeInstant.ts
```diff
@@ -0,0 +1,49 @@
...
+        return new Value(date, this);
+    }
+
+    toJsonValue(value: Value): string {
+        return value.isNull() ? null : value.getInstant().toString();
```
alansemenov:
```suggestion
        return value.isNull() ? null : this.valueToString(value);
```

src/main/resources/assets/admin/common/js/form/inputtype/time/DateTime.ts
```diff
@@ -31,7 +32,19 @@ export class DateTime
...
+
+        if (!defaultValue) {
+            return null;
+        }
+
+        if (DateTime.PATTERN.test(defaultValue)) {
+            return DateTimeUtil.fromString(defaultValue).toDate();
+        } else {
+            return RelativeTimeParser.parseToDateTime(defaultValue);
+        }
```
alansemenov:
```suggestion
        if (DateTime.PATTERN.test(defaultValue)) {
            return DateTimeUtil.fromString(defaultValue).toDate();
        }

        return RelativeTimeParser.parseToDateTime(defaultValue);
```

alansemenov (changes requested):
Some small things to fix

PR #4237, opened 2026-01-14
Missing WCAG attributes

alansemenov:


ashklianko:
Are these the elements Lighthouse complained on?

Issue #4371, opened 2026-03-24
SortableList: fix keyboard activation intercepting child button events


ashklianko:
When `fullRowDraggable` is enabled, pressing Enter or Space on interactive child elements (edit, remove buttons) triggers the drag activation instead of the button action. This affects all sortable lists including content selector, image selector, media selector, and the new site configurator input.

The root cause is in `SortableListItem`: dnd-kit's `listeners` (including `onKeyDown`) are spread onto the row div after the guarded `handleKeyDown` handler, overwriting it. The guard `e.target !== e.currentTarget` — which prevents drag activation when a child element is focused — never runs.

The fix strips `onKeyDown` from the spread listeners when `fullRowDraggable` is active, so the guarded handler always takes precedence.

#### References

`lib-admin-ui/src/main/resources/assets/admin/common/js/form2/components/sortable-list/index.tsx`

Example: ContentSelector in CS

<sub>*Drafted with AI assistance*</sub>

PR #4372, opened 2026-03-24
SortableList: fix keyboard activation intercepting child button event…

ashklianko:
…s #4371


edloidas (changes requested):
I have two problems with that fix:
- No stories to test and demonstrate that.
- This is actually not meant to have buttons initially. But since we already need them, to make navigation and tabulation work correctly, this must be a GridList, actually. If we handle drag only when focus is on drag handler and corresponding buttons when focus is on them, that should work correctly. Making a user to tab through all the X buttons and rows with tab is a real problem.

PR #4414, opened 2026-04-13
Impossible to type into input with regular expression #4411

### LINKED ISSUE #4411: Impossible to type into input with regular expression
In the Features project, create a content of type "Text Line with regular expression", where both inputs have regular expressions defined in the schema.

Bug: it's impossible to type anything into neither of the two fields.

skjulestad:
Fixed form2 TextLine regex handling so users can keep typing invalid intermediate values instead of the field clearing on each keystroke. The input now keeps local raw text, forwards that raw value upstream, and TextLineDescriptor validates rawValue when the stored value is null.

Added regression tests for all three layers: TextLineInput, InputField, and TextLineDescriptor.

### DIFF
--- a/src/main/resources/assets/admin/common/js/form2/components/text-line-input/TextLineInput.tsx
+++ b/src/main/resources/assets/admin/common/js/form2/components/text-line-input/TextLineInput.tsx
@@ -1,5 +1,6 @@
 import {Input} from '@enonic/ui';
 import type {JSX} from 'react';
+import {useEffect, useRef, useState} from 'react';
 
 import {ValueTypes} from '../../../data/ValueTypes';
 import type {TextLineConfig} from '../../descriptor';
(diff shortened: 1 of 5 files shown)


src/main/resources/assets/admin/common/js/form2/components/text-line-input/TextLineInput.tsx
```diff
@@ -9,32 +10,45 @@ import {Counter} from '../counter';
 
 const TEXT_LINE_INPUT_NAME = 'TextLineInput';
```
edloidas: Just a code style fix, but place the name right before the component. All that types and functions should be moved before it.

src/main/resources/assets/admin/common/js/form2/components/text-line-input/TextLineInput.tsx
```diff
@@ -9,32 +10,45 @@ import {Counter} from '../counter';
...
-}: InputTypeComponentProps<TextLineConfig>): JSX.Element => {
-    const stringValue = value.isNull() ? '' : (value.getString() ?? '');
+export type TextLineInputProps = InputTypeComponentProps<TextLineConfig>;
+
+function valueToString(value: TextLineInputProps['value']): string {
+    return value.isNull() ? '' : (value.getString() ?? '');
+}
+
+export const TextLineInput = ({value, onChange, onBlur, config, enabled, errors}: TextLineInputProps): JSX.Element => {
+    const [rawInput, setRawInput] = useState(() => valueToString(value));
```
edloidas, line 20: The code is correct. But it's not the things that you usually expect to see. I just want you to fully understand why you placed a function here, how it may affect performance, and why just calling `valueToString(value)` here wouldn't do much of a difference.
If you already fully understand this, then it's great, just ignore it.

edloidas (changes requested):
Looks great. Just a small style fix. Please, do amend and push.

ashklianko (approved):

PR #4460, opened 2026-05-05
No auto-grow on TextArea input type #4451

skjulestad:
TextArea now InputType should auto-grows.


alansemenov:
I remember there used to be an issue with this not working in Firefox. Is it still the case? @edloidas

PR #4535, opened 2026-06-09
Show server-side validation errors on form fields #4534

ashklianko:
- Added a ServerErrors context so consumers can feed content validation errors into form2. InputField now displays server errors per occurrence, ahead of client errors, and clears them on value edit (that occurrence) and on occurrence add/remove/move (the whole field). validateForm matches server errors by data path with occurrence indices and counts them toward validity — including optional fields — via a dedicated `server` flag on ValidationResult.


edloidas (changes requested):
You should really try using `code-cleanup` skill or something like that. CC just loves adding useless comments. And I can see lots of them here. 
Besides thet looks good

PR #4545, opened 2026-06-19
Incorrect TextArea behavior #4544

skjulestad:
Let text exceed the max limit with visual feedback as well as avoid clearing text when maxLength has been exceeded.


edloidas (approved):
Great work!

PR #4547, opened 2026-06-19
Long text overflows instead of wrapping in the TextArea #4546

skjulestad:
Add missing width control to avoid overflow within a grid.

### DIFF
--- a/src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
+++ b/src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
@@ -200,7 +200,7 @@ const OccurrenceListRoot = <C extends InputTypeConfig = InputTypeConfig>({
                 : errors.validationResults;
 
         return (
-            <div data-component={OCCURRENCE_LIST_NAME} className='grid gap-y-2'>
+            <div data-component={OCCURRENCE_LIST_NAME} className='grid min-w-0 gap-y-2'>
                 <InputLabel input={input} />
                 <Component
                     value={value}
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
```diff
@@ -260,7 +260,7 @@ const OccurrenceListRoot = <C extends InputTypeConfig = InputTypeConfig>({
         const showRemove = state.canRemove && state.values.length > 1;
 
         return (
-            <div data-component={OCCURRENCE_LIST_NAME} className='grid gap-y-2'>
+            <div data-component={OCCURRENCE_LIST_NAME} className='grid min-w-0 gap-y-2'>
```
ashklianko, line 263: The problem is only for single occurrence, you sure we need this here too?

src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
```diff
@@ -311,7 +311,7 @@ const OccurrenceListRoot = <C extends InputTypeConfig = InputTypeConfig>({
     }
 
     return (
-        <div data-component={OCCURRENCE_LIST_NAME} className='grid gap-y-2'>
+        <div data-component={OCCURRENCE_LIST_NAME} className='grid min-w-0 gap-y-2'>
```
ashklianko, line 314: The problem is only for single occurrence, you sure we need this here too?

ashklianko (changes requested):

src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
skjulestad: Yes, if not we keep min-w-auto default behavior.
Discovery: It might seem like textArea is wrapped in occurrence list even as a single component, to test make a new component text area input and inspect element.

src/main/resources/assets/admin/common/js/form2/components/occurrence-list/OccurrenceList.tsx
skjulestad: Yes, if not we keep min-w-auto default behavior.
Discovery: It might seem like textArea is wrapped in occurrence list even as a single component, to test make a new component text area input and inspect element.

PR #4562, opened 2026-06-23
Regression: tags autocompletion #4561

skjulestad:
Add tag autoComplete

### DIFF
--- a/src/main/resources/assets/admin/common/js/form2/components/input-field/InputField.tsx
+++ b/src/main/resources/assets/admin/common/js/form2/components/input-field/InputField.tsx
@@ -275,6 +275,7 @@ export const InputFieldResolved = ({
     ).toString();
 
     const fieldDataPath = fieldPath.startsWith('.') ? fieldPath.slice(1) : fieldPath;
+    const fieldPropertyPath = useMemo(() => PropertyPath.fromString(fieldPath), [fieldPath]);
     const serverEntries = serverErrors?.entries;
     const serverErrorsByOccurrence = useMemo(
         () => bucketServerErrorsByOccurrence(serverEntries ?? [], fieldDataPath),
@@ -619,6 +620,7 @@ export const InputFieldResolved = ({
                         occurrences={occurrences}
                         config={config}
                         input={input}
+                        dataPath={fieldPropertyPath}
                         enabled={enabled}
                         errors={displayValidation}
                     />
--- a/src/main/resources/assets/admin/common/js/form2/components/tag-input/TagInput.tsx
+++ b/src/main/resources/assets/admin/common/js/form2/components/tag-input/TagInput.tsx
@@ -12,17 +12,19 @@ import {
 import {rectSortingStrategy, SortableContext, sortableKeyboardCoordinates, useSortable} from '@dnd-kit/sortable';
 import {cn, IconButton, Input, Tooltip} from '@enonic/ui';
 import {GripVertical, X} from 'lucide-react';
-import {type JSX, type ReactElement, type RefObject, useEffect, useRef, useState} from 'react';
+import {type JSX, type ReactElement, type RefObject, useEffect, useId, useRef, useState} from 'react';
 
 import type {Value} from '../../../data/Value';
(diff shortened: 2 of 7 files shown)


src/main/resources/assets/admin/common/js/form2/components/tag-input/tagInputUtils.ts
```diff
@@ -36,6 +36,35 @@ export function hasTagLabel(values: Value[], label: string, excludedIndex?: numb
...
+        ) {
+            return result;
+        }
+
+        seenLabels.add(normalizedSuggestion);
+        result.push(normalizedSuggestion);
+        return result;
+    }, []);
+}
+
```
edloidas, line 71:
```suggestion
export function getSuggestedTagLabels(
    suggestions: string[],
    values: Value[],
    draft: string,
): string[] {
    const normalizedDraft = normalizeTagDraft(draft);

    if (!hasRenderableTagLabel(normalizedDraft)) {
        return [];
    }

    const unavailableLabels = new Set<string>([normalizedDraft]);

    for (const value of values) {
        const label = normalizeTagDraft(getTagLabel(value));

        if (hasRenderableTagLabel(label)) {
            unavailableLabels.add(label);
        }
    }

    const result: string[] = [];

    for (const suggestion of suggestions) {
        const label = normalizeTagDraft(suggestion);

        if (!hasRenderableTagLabel(label) || unavailableLabels.has(label)) {
            continue;
        }

        unavailableLabels.add(label);
        result.push(label);
    }

    return result;
}

```

src/main/resources/assets/admin/common/js/form2/components/tag-input/TagInput.tsx
```diff
@@ -69,7 +72,9 @@ type TagItemProps = {
     onRemove: () => void;
 };
 
-export type TagInputProps = SelfManagedComponentProps;
+export type TagInputProps = SelfManagedComponentProps & {
+    suggestTags?: (query: string) => Promise<string[]>;
+};
```
edloidas: I'd rather try to make it generic like this. Then in CS we can use it directly.
```
TagInputProps<C extends InputTypeConfig = InputTypeConfig> = SelfManagedComponentProps<C> & { suggestTags?: TagSuggester }
```

edloidas: Can present `TagSuggester` type as well.
```
export type TagSuggester = (query: string) => Promise<string[]>;
```

src/main/resources/assets/admin/common/js/form2/components/input-field/InputField.tsx
```diff
@@ -619,6 +619,7 @@ export const InputFieldResolved = ({
                         occurrences={occurrences}
                         config={config}
                         input={input}
+                        dataPath={PropertyPath.fromString(fieldPath)}
```
edloidas: You create new object all the time, and it will lead to component updates. Either wrap it with useMemo or just pass a string and convert it to object in function when it's really needed.

src/main/resources/assets/admin/common/js/form2/components/tag-input/tag-input.utils.ts
edloidas, line 1: Should be `tag-input.utils.ts`

src/main/resources/assets/admin/common/js/form2/components/tag-input/TagInput.tsx
```diff
@@ -689,6 +712,37 @@ export const TagInput = ({
...
+            setSuggestions([]);
+            setActiveSuggestionIndex(-1);
+            return undefined;
+        }
+
+        let stale = false;
+        setSuggestions([]);
+        setActiveSuggestionIndex(-1);
+
+        suggestTags(normalizedDraft).then(
```
edloidas: Should probably be debunced here.

edloidas (changes requested):

edloidas (approved):

PR #4601, opened 2026-07-17
Combobox validation fires immediately for a new content #4600

reisfmb:


alansemenov (changes requested):
We need to add tests as was done in https://github.com/enonic/lib-admin-ui/pull/4567/changes#diff-99dea1f3d08cf8becadcc0760a927dacdf307c5f424487b09ce6d9502074d44a

alansemenov (approved):

PR #4613, opened 2026-07-29
mobile friendly dnd #4606

skjulestad:
replace PointerSensor with separate mouse and touch sensors add delayed touch activation while preserving native scrolling remove touch-none from full-row sortable lists
prevent dragging when disabled in row and handle modes update dnd-kit sensor mocks

### DIFF
--- a/src/main/resources/assets/admin/common/js/form2/components/sortable-grid-list/SortableGridList.test.tsx
+++ b/src/main/resources/assets/admin/common/js/form2/components/sortable-grid-list/SortableGridList.test.tsx
@@ -66,6 +66,7 @@ const mocks = vi.hoisted(() => ({
     sortableContext: vi.fn(({children}: {children: unknown}) => children),
     cn: vi.fn((...tokens: Array<string | false | undefined>) => tokens.filter(Boolean).join(' ')),
     gripVertical: vi.fn(() => null),
+    mouseSensor: class MouseSensor {},
 }));
 
 vi.mock('react', () => ({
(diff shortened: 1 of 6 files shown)


ashklianko (changes requested):
1. Handle mode gets a 200 ms press-and-hold — and it's the only mode our consumers use. The activation constraint lives on DndContext, so the delay applies to grip-handle drags too. On touch, grabbing the grip and dragging immediately moves >5 px inside the delay window, so tolerance cancels and the page scrolls instead. Both in-repo consumers — OccurrenceList.tsx:270, ComboBoxInput.tsx:138 — render SortableGridList without fullRowDraggable; SortableList has no in-repo consumers beyond stories. So the touch-none removal only affects a mode nothing here uses, while every form now needs a deliberate hold on a ~20 px button. Fix: conditional constraint (fullRowDraggable ?{delay: 200, tolerance: 5} : {distance: 5}) plus touch-none on the grip button, or bypassActivationConstraint.

2. Middle-click now starts a drag. PointerSensor rejected !isPrimary || button !== 0; MouseSensor only rejects button === 2. Button 1 passes, and it collides with browser middle-click autoscroll.

3. Add a test pinning the sensor configuration, plus an enabled: false case — enabled appears once in the test file, so the disabled: !enabled || !isMovable fix is unverified, and the mock wouldn't have caught the swap either (useSensor(undefined, opts) stays green).
4. Some small things:
- Most new enabled guards are dead code (SortableList.tsx:225,231,279; SortableGridList.tsx:408,415,494) — useDraggable already returns listeners: undefined when disabled. In SortableList they were dead on arrival (master had it from #4602).The className guards are real and should stay.
- select-none parity — SortableGridList.tsx:475 lacks what SortableList.tsx:314 has; consistency, not a live bug.
-  Consider touch-manipulation instead of leaving touch-action: auto.
- TagInput.tsx:687 still on PointerSensor — same breakage, follow-up issue?
- Pre-existing: inline useSensor options literals invalidate the whole listeners chain every render — that's what makes the useMemo at SortableList.tsx:231 a no-op.
-  Minor: delay: 200 is short-ish; pen input is now implicit rather than explicit.

ashklianko (approved):

